PR #1, opened 2018-01-30
test squeleton

antoine-de:


antoine-de:
sorry for this PR, I screwd up and opened it in the wrong repository 🙄 

FYI, we're trying to see if writing [rust-geo](https://github.com/georust/rust-geo) bindings will be easy (to have easily all geos algorithm in rust-geo)

mthh:
No worries for the PR!

I used this project to play a bit with rust (and because I possibly needed algorithm proposed by GEOS). But I prefer to warn you that I did not really continue the development of this package notably because of an error that I could not avoid: a segmentation error occurs when an invalid WKT geometry is created (I thought to ask for help on users.rust-lang.org but I never took the time, anyway, glad to see that there is interest for GEOS in rust!)

PR #4, opened 2018-04-03
change upstream

antoine-de:
tiddy up before releasing the crates on crates.io


mthh:
Great that it will soon be ready to be published on crates.io!

PR #6, opened 2018-04-03
Add description and pin version

antoine-de:
last step before pushing to crates.io

we need pinned dependencies and a description

Sorry for all those PR, I'm not a crates.io expert :stuck_out_tongue:


antoine-de:
I tested the deployment (in order to stop creating additional PR :smile: )

so we now have a crates.io crate :tada: 

https://crates.io/crates/geos

I had to update the version number to be able to reupdate the crates after the first mistakes :confused:

PR #7, opened 2018-04-03
uses Results everywhere

antoine-de:
lots of geos operation can fail so we wrap those in Result to remove every `panic!`

I bumped the version number as this breaks the api

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ This project is checked with valgrind, but if you stumble on a crash feel free t
 
 ### Usage example
 
-You can check the examples in the `examples/` directorry.
+You can check the examples in the `examples/` directory.
 
 ### Constructing geometries from WKT:
 
--- a/README.md
+++ b/README.md
@@ -14,78 +14,27 @@ This project is checked with valgrind, but if you stumble on a crash feel free t
 
 ### Usage example
 
+You can check the examples in the `examples/` directorry.
+
 ### Constructing geometries from WKT:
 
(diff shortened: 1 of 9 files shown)


src/ffi.rs
```diff
@@ -213,8 +214,12 @@ impl CoordSeq {
...
-    pub fn new_from_c_obj(c_obj: *mut GEOSCoordSequence) -> CoordSeq {
-        CoordSeq(c_obj)
+    pub fn new_from_c_obj(c_obj: *mut GEOSCoordSequence) -> GeosResult<CoordSeq> {
+        if c_obj.is_null() {
+            Err(Error::NoConstructionFromNullPtr)
+        } else {
+            Ok(CoordSeq(c_obj))
+        }
     }
     pub fn set_x(&self, idx: u32, val: f64) -> i32 {
```
mthh: Given that it may also fail, should not these 3 methods (`set_x`, `set_y` and  `set_z`) also return a `Result`? (rather than letting the user look at the return value)

antoine-de: oh right, I missed those, thanks !

README.md
```diff
@@ -14,110 +14,54 @@ This project is checked with valgrind, but if you stumble on a crash feel free t
 
 ### Usage example
 
+You can check the examples in the `examples/` directorry.
```
amatissart: "directorry" :arrow_right: "directory"

PR #8, opened 2018-04-05
minor cleanup: remove area and _type from Geom

antoine-de:
it was not mandatory to store them in the geometry, they can be computed on the fly to we remove those 2 fields.

And use the enum GeometryType not to carry i32 around.


antoine-de:
yes I have the rights but it's nice for another to review them, thank you very much !

PR #9, opened 2018-04-10
Fix CoordSeq bug

antoine-de:
A closed ring with only 2 points was crashing libgeos

the ring: [p1, p2, p1] is not valid for geos (and it thus crashes without any nice explanation).

We use the same approach as [shapely](https://shapely.readthedocs.io) a quite used geos python wrapper and  
* we close the unclosed `LinearRing`
* we consider the 2 points closed ring as a 3 points unclosed ring (with the 2 last points beeing at the same position :laughing: ) and thus close it too.

Since it was the last blocking bug for us, I propose to bump the crate version to `1.0.0` to express the fact that it will soon be used in production somewhere :stuck_out_tongue_winking_eye:

### DIFF
--- a/src/ffi.rs
+++ b/src/ffi.rs
@@ -407,7 +407,7 @@ impl GGeom {
 
     /// get the underlying geos CoordSeq object from the geometry
     ///
-    /// Note: this clones the underlying CoordSeq not avoid double free
+    /// Note: this clones the underlying CoordSeq to avoid double free
     /// (because CoordSeq handles the object ptr and the CoordSeq is still owned by the geos geometry)
     /// if this method's performance becomes a bottleneck, feel free to open an issue, we could skip this clone with cleaner code
     pub fn get_coord_seq(&self) -> Result<CoordSeq, Error> {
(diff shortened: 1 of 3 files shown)


src/ffi.rs
```diff
@@ -389,11 +405,16 @@ impl GGeom {
         return if rv == 1 { true } else { false };
     }
 
+    /// get the underlying geos CoordSeq object from the geometry
+    ///
+    /// Note: this clones the underlying CoordSeq not avoid double free
```
amatissart: typo in comment: "to avoid double free" ?

src/from_geo.rs
```diff
@@ -35,21 +40,28 @@ impl<'a> TryInto<GGeom> for &'a LineString<f64> {
...
+/// Convert a geo::LineString to a geos LinearRing
+/// a LinearRing should be closed so cloase the geometry if needed
 impl<'a> TryInto<GGeom> for &'a LineRing<'a> {
     type Err = Error;
 
     fn try_into(self) -> Result<GGeom, Self::Err> {
         let points = &(self.0).0;
-        let coord_seq = create_coord_seq(&points)?;
+        let nb_points = points.len();
+        if nb_points >= 1 && nb_points < 3 {
```
amatissart: Wouldn't it be clearer to write `if nb_points < 3` ? And that would simplify the other conditions (lines 56,58).  
What is supposed to happen if nb_points <= 1 ?

antoine-de: we want an empty `LinearRing` :cry: 

so I don't think we can cleanup this

PR #11, opened 2018-04-11
Return a GeosResult on the various binary predicate methods

### LINKED ISSUE #10: Use Result on binary predicates (to handle possible GEOS exception)
In order to expose a safe interface and to not alter the results returned by GEOS I guess we should check the value returned by the various [binary predicates](https://github.com/OSGeo/geos/blob/master/capi/geos_c.h.in#L1668) and wrap it in a `Result`.
GEOS is returning 2 for an exception (that we could transform to an `Err`) or 1/0 if its true/false (that we could wrap in a `Ok`).
Currently we are doing like this : https://github.com/georust/rust-geos/blob/master/src/ffi.rs#L464

If no one is already working on it I will send a PR today, sorry this error remained in the code so long of my fault!

mthh:
This PR wraps the result of the various binary predicates on `GGeom` and `PreparedGGeom` in a `GeosResult` in order to fix #10 .

I wanted to include a test case but I can't figure out how to trigger such an exception for now.

It bumps the version number to 1.1.0 as it breaks the API.

### DIFF
--- a/src/ffi.rs
+++ b/src/ffi.rs
@@ -2,7 +2,7 @@ use libc::{atexit, c_char, c_double, c_int, c_uint, c_void, size_t};
 use std::sync::{Once, ONCE_INIT};
 use std::ffi::{CStr, CString};
 use std::{ptr, str};
-use error::{Error, Result as GeosResult};
+use error::{Error, Result as GeosResult, PredicateType};
 use std;
 use num_traits::FromPrimitive;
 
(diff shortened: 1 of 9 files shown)


src/ffi.rs
```diff
@@ -453,77 +453,149 @@ impl GGeom {
         result
     }
 
-    pub fn is_ring(&self) -> bool {
+    pub fn is_ring(&self) -> GeosResult<bool> {
         let rv = unsafe { GEOSisRing(self.c_obj()) };
-        return if rv == 1 { true } else { false };
+        if rv == 1 {
```
antoine-de: since the pattern can be found several times, maybe we can wrap this in a method:


```rust
fn check_geos_predicate<F>(val: i32, on_error: F) -> GeosResult<bool>
where
    F: FnOnce() -> Error {
    match val {
        1 => Ok(true),
        0 => Ok(false),
        _ => Err(on_error())
    } 
}
```

I'm not sure about the error function, but this way the error (and the string) is created only on error ([clippy](https://github.com/rust-lang-nursery/rust-clippy) prefer `Option::ok_or_else` to `Option::ok_or` for this reason)

mthh: Sure I guess it will be really more clean. I'm not sure about the error function either, especially if we want to keep the name of the function that triggers the error.

mthh: I made a `check_geos_predicate` function which take the return value and an EnumVariant. This way I guess the error and the string are only created on error as you suggest (but I guess it's not really elegant to match on this new big Enum  in order to create the error), so don't hesitate to suggest me a better idea!

antoine-de: you could just have kept the error as before no ?
```rust
    pub fn is_ring(&self) -> GeosResult<bool> {
        let rv = unsafe { GEOSisRing(self.c_obj()) };
        check_geos_predicate(rv, || Error::GeosError("computing isRing".into()))
    }
```

but your solution is interesting as it removes the string.

It may be overkill, but maybe we could leverage the [Failure](https://github.com/rust-lang-nursery/failure) crate for this ?

I pushed a [toy implementation](https://github.com/antoine-de/rust-geos/commit/ce9794510a6ff8fb1cee7ece25cd28e9eeba3fed)

honestly I'm not convinced that the complexity is worth it compared to the initial version with a lambda for the error. What do you think ?

I added some unit test that could be nice to get though.

mthh: Excepting this big PredicateType Enum, I find your solution pretty elegant, so I changed the code to follow it. Sorry for the lack of imagination of mine to tackle this issue.

This is a detail but as we are referring to 'libgeos method' in the error message, i don't think it really makes sense to transform the Enum member to snake case as they match GEOS function names in camel case (?)

antoine-de: ok, it's up to you, both are fine for me

mthh:
@antoine-de Sorry, I planned to publish this new release on crates.io but I do not seem to have the necessary rights. Any advice ?

antoine-de:
I just added the all of the [GeoRust/core](https://github.com/orgs/georust/teams/core) team to the crates.io owner, should work now, sorry :confused: 

However I published the crates.io yesterday before reading the crates.io documentation, so we need to wait for another PR to test if you can publish a version.

btw not related at all, but we'll go to the [state of the map france](https://wiki.openstreetmap.org/wiki/FR:State_of_the_Map_France_2018) this year (to explain why we needed `rust-geos` :stuck_out_tongue_winking_eye: ), will you go there to ?

mthh:
Oh great for SOTM-fr, I will be there too! (only from Friday night I think, because of other commitments + I didn't submit anything this year). For sure I will not miss your intervention if it is Saturday (otherwise let's just grab a coffee there!)

PR #15, opened 2018-04-26
ffi cleanups

### LINKED ISSUE #13: Some unsafe functions are not tagged unsafe
Any function that dereference a raw pointer, or allow to, later, dereference a raw pointer, must be tagged unsafe. As a side effect, almost all the function taking a raw pointer as a parameter must be tagged unsafe.

Some such functions:
 - `GGeom::new_from_c_obj`
 - `CoordSeq::new_from_c_obj`

TeXitoi:
As a breaking change, `*::new_from_c_obj` are now private.

Use clean unsafe tagging and conventions.

Fix #13

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "geos"
-version = "1.1.0"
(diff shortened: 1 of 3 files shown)


antoine-de:
the version 2.0.0 has been publised on crates.io

PR #24, opened 2019-03-10
Add UnaryUnion and Union

crutt:


mthh:
Thanks !
Unless someone have an opinion on the fact of not using `unary_union` on `self` but on a geometry given in argument, it looks good to me !

src/ffi.rs
```diff
@@ -628,6 +636,12 @@ impl GGeom {
         Ok(obj)
     }
 
+    pub fn unary_union(g: &GGeom) -> GeosResult<GGeom> {
```
antoine-de: Sorry for being late, I do aggre with @mthh it's minor, but it might be better to use self instead of another `GGeom` no ?

If that's fine for you, I think I'll change it.

mthh: It's fine for me if you change it.
I did not release a version after merging it (i was thinking to some other minors changes, like #25, and maybe adding some documentation, but I ran out of time this week).
I mean, `unary_union` taking `self` or  a `ggeom` as argument does not shock me, but I thought it would be even better if the methods will be somehow documented.

PR #26, opened 2019-03-17
take self for unary union, for coherence with the other methods

antoine-de:


mthh:
If I'm not mistaken UnaryUnion works notably on geometrycollections. Maybe we should also expose a function that makes it possible to create a geometrycollection from a `Vec` of `GGeom` ?

mthh:
It's probably better to open new issues and discuss them if needed than discuss them here (so I opened #27 and #28). My bad, it's merged now :)

PR #30, opened 2019-03-22
Add create_multilinestring and create_geometrycollection

### LINKED ISSUE #27: Adding methods to create MultiLineString and GeometryCollection
We currently have methods to create MultiPoint (https://github.com/georust/geos/blob/master/src/ffi.rs#L600) and MultiPolygon (https://github.com/georust/geos/blob/master/src/ffi.rs#L581) from a `Vec` of `GGeom` but it seems it lacks methods for construction of MultiLineString and GeometryCollection from a `Vec` of `GGeom`.

### LINKED ISSUE #29: Invalid memory reference if all geometries aren't of same expected type on create_multi_*
Step to reproduce :

```rust
let vec_geoms = vec![
    GGeom::new("POINT (1.3 2.4)").unwrap(),
    GGeom::new("LINESTRING(1 1,10 50,20 25)").unwrap(),
];
let multi_point = GGeom::create_multipoint(vec_geoms)?;
```

mthh:
This PR adds `create_multilinestring` and `create_geometrycollection` methods (fixes #27) and forces to check the type of each geometry given in argument in `create_multipoint`, `create_multilinestring` and `create_multipolygon` methods (fixes #29).

### DIFF
--- a/src/ffi.rs
+++ b/src/ffi.rs
@@ -773,14 +773,14 @@ fn check_geos_predicate(val: i32, p: PredicateType) -> GeosResult<bool> {
 
 fn check_same_geometry_type(geoms: &[GGeom], geom_type: GEOSGeomTypes) -> bool {
     geoms.iter()
-        .map(|g|
-            if g.geometry_type()? == geom_type {
-                Ok(true)
+        .all(|g| {
+            let t = g.geometry_type();
+            if t.is_err() {
+                false
             } else {
-                Err(Error::InvalidGeometry("".to_string()))
+                t.unwrap() == geom_type
+            }
         })
-        .collect::<GeosResult<Vec<bool>>>()
-        .is_ok()
 }
 
 fn create_multi_geom(mut geoms: Vec<GGeom>, output_type: GEOSGeomTypes) -> GeosResult<GGeom> {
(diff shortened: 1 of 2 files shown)


src/ffi.rs
```diff
@@ -778,6 +771,37 @@ fn check_geos_predicate(val: i32, p: PredicateType) -> GeosResult<bool> {
     }
 }
 
+fn check_same_geometry_type(geoms: &[GGeom], geom_type: GEOSGeomTypes) -> bool {
+    geoms.iter()
+        .map(|g|
```
antoine-de: I think you can rewrite this like:

```rust
geoms.iter().all(|g| g.geometry_type == Ok(geom_type))
```

mthh: Ah great :+1:  I was looking for some short-circuiting method like `all` (so I go with `collect` and it ended up a bit messy)

mthh: I somehow changed it but it can't be as clean as in your example because we still have to verify that `geometry_type` does not return an error.

antoine-de: oh you're right, it seems that the `Error` is not comparable sorry :roll_eyes:  

It might be a bit more rustic to write this:
```rust
    geoms
        .iter()
        .all(|g| g.geometry_type().map(|t| t == geom_type).unwrap_or(false))
```

mthh: Respect :clap:  I struggled with the various `unwrap_*` and `map_*` for a while and I couldn't find a clean solution!

PR #33, opened 2019-03-27
Start binding GEOSContextHandle type

GuillaumeGomez:


mthh:
Oh great ! I really had troubles when I tried to bind `GEOSContext_setNoticeMessageHandler_r` / `GEOSContext_setErrorMessageHandler_r`. I will learn from your example for sure !

GuillaumeGomez:
I just thought that those two functions could have closures will the same lifetime as the object instead of `'static`. That could be interesting.

GuillaumeGomez:
Oh and travis has some weird lazy failures...

src/ffi.rs
```diff
@@ -1,35 +1,41 @@
...
 #[repr(C)]
-struct GEOSCoordSequence { private: [u8; 0]}
+struct GEOSCoordSequence { private: [u8; 0] }
 #[repr(C)]
-struct GEOSGeometry { private: [u8; 0]}
+struct GEOSGeometry { private: [u8; 0] }
+#[repr(C)]
+struct GEOSContextHandle_HS { private: [u8; 0] }
+#[allow(non_camel_case_types)]
+type GEOSContextHandle_t = *mut GEOSContextHandle_HS;
```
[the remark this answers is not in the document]
GuillaumeGomez: We could but I'm not a big fan of it. :)

GuillaumeGomez:
Okay, should be ready now!

src/ffi.rs
```diff
@@ -142,22 +150,379 @@ extern "C" {
...
+    /// };
+    /// let point_geom = GGeom::new("POINT (2.5 2.5)").expect("Invalid geometry");
+    /// let wkb_buf = context_handle.geom_to_WKB_buf(&point_geom)
+    ///                             .expect("conversion to WKB failed");
+    /// let new_geom = context_handle.geom_from_WKB_buf(wkb_buf.as_ref())
+    ///                              .expect("conversion to WKB failed");
+    /// assert!(point_geom.equals(&new_geom) == Ok(true));
+    /// ```
+    #[allow(non_snake_case)]
+    pub fn geom_to_WKB_buf(&self, g: &GGeom) -> Option<CVec<u8>> {
```
GuillaumeGomez: I just thought: should I return a `String` in here?

src/ffi.rs
```diff
@@ -142,22 +150,379 @@ extern "C" {
...
+    /// };
+    /// let point_geom = GGeom::new("POINT (2.5 2.5)").expect("Invalid geometry");
+    /// let wkb_buf = context_handle.geom_to_WKB_buf(&point_geom)
+    ///                             .expect("conversion to WKB failed");
+    /// let new_geom = context_handle.geom_from_WKB_buf(wkb_buf.as_ref())
+    ///                              .expect("conversion to WKB failed");
+    /// assert!(point_geom.equals(&new_geom) == Ok(true));
+    /// ```
+    #[allow(non_snake_case)]
+    pub fn geom_from_WKB_buf(&self, wkb: &[u8]) -> GeosResult<GGeom> {
```
GuillaumeGomez: I just thought: should I take a `String` as parameter in here?

src/ffi.rs
```diff
@@ -142,22 +150,379 @@ extern "C" {
...
+    /// };
+    /// let point_geom = GGeom::new("POINT (2.5 2.5)").expect("Invalid geometry");
+    /// let wkb_buf = context_handle.geom_to_HEX_buf(&point_geom)
+    ///                             .expect("conversion to WKB failed");
+    /// let new_geom = context_handle.geom_from_HEX_buf(wkb_buf.as_ref())
+    ///                              .expect("conversion to WKB failed");
+    /// assert!(point_geom.equals(&new_geom) == Ok(true));
+    /// ```
+    #[allow(non_snake_case)]
+    pub fn geom_from_HEX_buf(&self, hex: &[u8]) -> GeosResult<GGeom> {
```
GuillaumeGomez: I just thought: should I take a `String` as parameter in here?

src/ffi.rs
```diff
@@ -142,22 +150,379 @@ extern "C" {
...
+    /// };
+    /// let point_geom = GGeom::new("POINT (2.5 2.5)").expect("Invalid geometry");
+    /// let wkb_buf = context_handle.geom_to_HEX_buf(&point_geom)
+    ///                             .expect("conversion to WKB failed");
+    /// let new_geom = context_handle.geom_from_HEX_buf(wkb_buf.as_ref())
+    ///                              .expect("conversion to WKB failed");
+    /// assert!(point_geom.equals(&new_geom) == Ok(true));
+    /// ```
+    #[allow(non_snake_case)]
+    pub fn geom_to_HEX_buf(&self, g: &GGeom) -> Option<CVec<u8>> {
```
GuillaumeGomez: I just thought: should I return a `String` in here?

src/ffi.rs
TeXitoi: WKB is binary, not utf-8, so `&[u8]` is your only solution.

TeXitoi: Why WKB in caps? Why not `geom_from_wkb_buf` as used in the rust convension? https://doc.rust-lang.org/nightly/std/str/fn.from_utf8.html

src/ffi.rs
```diff
@@ -142,22 +150,379 @@ extern "C" {
...
+    ///                              .expect("conversion to WKB failed");
+    /// assert!(point_geom.equals(&new_geom) == Ok(true));
+    /// ```
+    #[allow(non_snake_case)]
+    pub fn geom_to_WKB_buf(&self, g: &GGeom) -> Option<CVec<u8>> {
+        let mut size = 0;
+        unsafe {
+            let ptr = GEOSGeomToWKB_buf_r(self.ptr, g.as_raw(), &mut size);
+            if ptr.is_null() || size == 0 {
+                None
```
TeXitoi: If ptr is not null but size=0, you will leak, no?

src/ffi.rs
```diff
@@ -142,22 +150,379 @@ extern "C" {
...
+    ///
+    /// let context_handle = match GContextHandle::init() {
+    ///     Ok(context_handle) => context_handle,
+    ///     Err(e) => panic!("An error occurred: {}", e),
+    /// };
+    /// let point_geom = GGeom::new("POINT (2.5 2.5)").expect("Invalid geometry");
+    /// let wkb_buf = context_handle.geom_to_HEX_buf(&point_geom)
+    ///                             .expect("conversion to WKB failed");
+    /// let new_geom = context_handle.geom_from_HEX_buf(wkb_buf.as_ref())
+    ///                              .expect("conversion to WKB failed");
```
TeXitoi: bad expect

src/ffi.rs
GuillaumeGomez: I can update the name. :)

src/ffi.rs
```diff
@@ -142,22 +150,379 @@ extern "C" {
...
+impl Drop for GContextHandle {
+    fn drop<'a>(&'a mut self) {
+        unsafe { GEOS_finish_r(self.ptr) };
+
+        let previous_ptr = self.notice_message.replace(::std::ptr::null_mut());
+        if !previous_ptr.is_null() {
+            // We free the previous closure.
+            let _callback: Box<Option<Box<dyn Fn(&str) + 'a>>> =
+                unsafe { Box::from_raw(previous_ptr as *mut _) };
+        }
```
TeXitoi: This will mix badly if we use a C lib that also use geos, no?

src/ffi.rs
GuillaumeGomez: I'm not sure such a case is actually possible...

src/ffi.rs
GuillaumeGomez: Unless you call this function on the same object but without the binding, it won't. It's not a global function after all.

src/ffi.rs
GuillaumeGomez: Ah right: I mixed up the name with WKT.

src/ffi.rs
TeXitoi: as `free` a null ptr is valid, why not just returning a `CVec<u8>` and don't care of the empty case?

GuillaumeGomez: Fine by me.

GuillaumeGomez:
Anything else I should change?

src/ffi.rs
```diff
@@ -142,22 +150,379 @@ extern "C" {
...
+    }
+
+    /// Sets WKB output dimensions.
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// use geos::{GContextHandle, Dimensions};
+    ///
+    /// let context_handle = match GContextHandle::init() {
```
antoine-de: by curiosity, why don't you `.expect` here ?

GuillaumeGomez: More easy to read but that's just my point of view. I can change it if you prefer?

antoine-de: nop, I was just wondering, both are fine

GuillaumeGomez:
Updated.

PR #38, opened 2019-03-30
Implement ContextHandle inside GGeom directly

GuillaumeGomez:
@TeXitoi Does the API looks better this way? If so, we'll have to add a lot more examples in the docs stating that setting `GContextHandle` inside an object is better than just calling the object.

### DIFF
--- a/src/context_handle.rs
+++ b/src/context_handle.rs
@@ -1,5 +1,3 @@
-use crate::GGeom;
-use c_vec::CVec;
 use enums::{ByteOrder, Dimensions};
 use error::{Error, GResult};
 use ffi::*;
@@ -37,7 +35,7 @@ macro_rules! set_callbacks {
 set_callbacks!(GEOSContext_setNoticeMessageHandler_r, set_notif, notif_callback, last_notification);
 set_callbacks!(GEOSContext_setErrorMessageHandler_r, set_error, error_callback, last_error);
 
-struct PtrWrap<T>(T);
+pub(crate) struct PtrWrap<T>(pub T);
 
 impl<T> Deref for PtrWrap<T> {
     type Target = T;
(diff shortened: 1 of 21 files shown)


GuillaumeGomez:
After discussion, I'll change the following things:

 * No more usage of functions which not use `HandleContext` (the ones with name which doesn't end with "_r") because they're not thread-safe.
 * All objects will now have a `HandleContext` (in the object or given when calling the method, not clear yet, to be tested).
 * Save X last errors/notifications so the user can take a look at them whenever they want.

src/context_handle.rs
```diff
@@ -31,14 +74,36 @@ impl GContextHandle {
         initialize();
```
TeXitoi: initialize is not needed as it doesn't have `_t`

GuillaumeGomez: Good point. I'll remove it everywhere.

src/coord_seq.rs
```diff
@@ -1,105 +1,254 @@
...
+    ///                               .expect("failed to create CoordSeq");
+    /// let positions2: &[(f64, f64)] = &[(0., 0.), (1., 2.), (14., 5.)];
+    /// for (pos, (x, y)) in positions2.into_iter().enumerate() {
+    ///     coord_seq2.set_x(pos, *x).expect("failed to set x...");
+    ///     coord_seq2.set_y(pos, *y).expect("failed to set y...");
+    /// }
+    /// assert!(coord_seq2.get_x(1) == Ok(1.));
+    /// ```
+    pub fn new(size: u32, dims: CoordDimensions) -> GResult<CoordSeq<'a>> {
         initialize();
```
TeXitoi: I say nooo, nooo, nooo, nooo ;-)

GuillaumeGomez:
Well, that's why I was against this but the user can wrap it pretty easily by using channels.

GuillaumeGomez:
I removed `initialize` calls.

GuillaumeGomez:
Beta version of rust has memory leaks on threads apparently. That's pretty funny. :3

PR #39, opened 2019-04-01
Add missing parameter to compute_voronoi function

GuillaumeGomez:


GuillaumeGomez:
Considering there isn't much changes in here, I'll just go ahead and merge it since I kinda need it. :)

PR #42, opened 2019-04-11
Improvements

GuillaumeGomez:


src/coord_seq.rs
```diff
@@ -53,6 +61,80 @@ impl<'a> CoordSeq<'a> {
...
+
+        if size > 0 {
+            let dims = data[0].len();
+            if let Err(e) = CoordDimensions::try_from(dims as _) {
+                return Err(Error::GenericError(e.to_owned()));
+            }
+            if !data.iter().skip(1).all(|x| x.len() == dims) {
+                return Err(Error::GenericError("All vec entries must have the same size!".to_owned()));
+            }
+            match match GContextHandle::init() {
```
antoine-de: wouldn't it be more straighforward to write this with a `and_then` ?

I find the `match match` a bit difficult to read

src/coord_seq.rs
```diff
@@ -53,6 +61,80 @@ impl<'a> CoordSeq<'a> {
...
+                Ok(coord) => {
+                    let raw_context = coord.get_raw_context();
+                    let raw_coord = coord.as_raw();
+
+                    let funcs = [GEOSCoordSeq_setX_r,
+                                 GEOSCoordSeq_setY_r,
+                                 GEOSCoordSeq_setZ_r];
+
+                    for (line, line_data) in data.iter().enumerate() {
+                        for (pos, elem) in line_data.iter().enumerate() {
```
antoine-de: note sure about this, but shouldn't you check than there are never more than 3 elements ?

src/coord_seq.rs
```diff
@@ -162,6 +339,42 @@ impl<'a> CoordSeq<'a> {
...
+    ///
+    /// let coords = CoordSeq::new(2, CoordDimensions::ThreeD)
+    ///                       .expect("failed to create CoordSeq");
+    /// assert!(coords.number_of_lines() == Ok(2));
+    ///
+    /// let coords = CoordSeq::new_from_vec(&[&[1.], &[2.], &[3.], &[4.]])
+    ///                       .expect("failed to create CoordSeq");
+    /// assert!(coords.number_of_lines() == Ok(4));
+    /// ```
+    pub fn number_of_lines(&self) -> GResult<usize> {
```
antoine-de: not sure about this `line` naming.
line can be quite a lot of things in the `geo` context.

why don't you like `size()` (and `idx` instead of `line` in the signature/comments) ?

src/coord_seq.rs
GuillaumeGomez: It's done before when I check the dimensions of each line.

src/coord_seq.rs
GuillaumeGomez: Because you can see the `CoordSequence` as a matrix, therefore it's easier to speak about lines rather than "index" which is very vague.

src/coord_seq.rs
GuillaumeGomez: Good suggestion! I just like chaining `match` but you're right!

GuillaumeGomez:
Updated.

src/coord_seq.rs
```diff
@@ -96,7 +97,7 @@ impl<'a> CoordSeq<'a> {
             if !data.iter().skip(1).all(|x| x.len() == dims) {
                 return Err(Error::GenericError("All vec entries must have the same size!".to_owned()));
             }
-            match match GContextHandle::init() {
+            match GContextHandle::init() {
```
antoine-de: can't you put and_then on the first one too ?

`GContextHandle::init().and_then( unsafe {...}).and_then(|coord|...`

GuillaumeGomez: You're asking a bit too much of me! I love `match`es. :'(

TeXitoi:
I'll rewrote all of that with `map_err` and `?` personally.

Le 11 avril 2019 16:49:13 CEST, Guillaume Gomez <notifications@github.com> a écrit :
>GuillaumeGomez commented on this pull request.
>
>
>
>> @@ -96,7 +97,7 @@ impl<'a> CoordSeq<'a> {
>             if !data.iter().skip(1).all(|x| x.len() == dims) {
>return Err(Error::GenericError("All vec entries must have the same
>size!".to_owned()));
>             }
>-            match match GContextHandle::init() {
>+            match GContextHandle::init() {
>
>You're asking a bit too much of me! I love `match`es. :'(
>
>-- 
>You are receiving this because you are subscribed to this thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/georust/geos/pull/42#discussion_r274462889

-- 
Envoyé de mon appareil Android avec Courriel K-9 Mail. Veuillez excuser ma brièveté.

GuillaumeGomez:
This PR is big enough as is. Next step: splitting FFI part from this crate.

@TeXitoi: I'm not a big fan of that but if you want it, please open a PR. :)

mthh:
@GuillaumeGomez I havent been able to follow your PRs very closely but thank you for all the work in improving that library !

TeXitoi:
I dont really care that's just style.

GuillaumeGomez:
@mthh No problem and you're welcome! ;)

GuillaumeGomez:
@TeXitoi Style is important so feel free to apply yours anytime. :p

PR #43, opened 2019-04-12
Split out FFI part and prevent rebuilds

GuillaumeGomez:


GuillaumeGomez:
The first commit splits out the FFI part into another crate (which is available in https://github.com/georust/geos-sys).

The second commit allows to prevent rebuild by generating a doc comment from the README instead of generating a test file everytime.

antoine-de:
wouldn't it have been easier to maintain if both crates were in the same repository ?

GuillaumeGomez:
That's up to debate. I personally prefer multi-repositories setup.

PR #73, opened 2019-10-10
Allow conversion from/to GeoJSON

mthh:
This PR propose to add a `json` feature flag to handle the conversion from / to [geojson](https://github.com/georust/geojson) crate (instead of having to create [geo-type](https://github.com/georust/geo) geometries - themselves relying on WKT - to do it).


GuillaumeGomez:
Thanks! Completely forgot about this PR, sorry...

PR #74, opened 2019-11-16
feat: Add basic STRtree wrapper

gschulze:
Basic STRtree wrapper with insert and query methods. I'm fairly new to Rust, so I wasn't sure about the correct way to store (references to) items in the tree. The proposed implementation introduces a generic type parameter <I> corresponding to the type of items that can be inserted into the tree. Internally, items are boxed and the raw pointer is passed to the GEOS implementation. When the tree is dropped, the implementation ensures that those pointers are freed.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -52,6 +52,10 @@ pub use geometry::{
 pub use prepared_geometry::{
     PreparedGeometry,
 };
+pub use spatial_index:: {
+    SpatialIndex,
+    STRtree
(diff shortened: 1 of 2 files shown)


GuillaumeGomez:
Sorry, didn't receive the notification until @amatissart requested a review from me... Thanks a lot for the PR!

gschulze:
You're welcome :) Thanks for merging!

PR #75, opened 2020-01-09
Fixes bug in armv7 build

thomasantony:
Hardcoding the pointer type as `*const i8` causes the build to fail on the `armv7` platform and possibly other ARM-based targets as well. Changing this to `c_char` is a more general solution that makes `geos` compilable for multiple target architectures.


GuillaumeGomez:
Thanks!

PR #77, opened 2020-07-23
Update to geos-sys 2.0 and fix multiple issues

GuillaumeGomez:
* Update to geos-sys 2.0
* Strenghten checks on *const/*mut usage by creating a new AsRawMut trait
* Create a new ConstGeometry type to better handle the split with Geometry

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: rust
 
-dist: xenial
+dist: groovy
 
 rust:
   - stable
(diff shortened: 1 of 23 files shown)


GuillaumeGomez:
*sigh*

I really need to use github-actions and remove travis, it's just too annoying. For now let's merge it as is I guess...

GuillaumeGomez:
Weird error on beta. Let's ignore it for now and **finally** merge.

PR #87, opened 2021-06-10
add known broken test for dropping geom used in prepared geometry

gauteh:


GuillaumeGomez:
Writing the fix. Merging to make my work easier. Thanks again!

PR #88, opened 2021-06-10
Fixed lifetime issues on PreparedGeometry

### LINKED ISSUE #86: create a PreparedGeometry<'static> from a Geometry<'static> and dropping the latter makes the first unusable
Seems like the PreparedGeometry should have a lifetime dependency on the original Geometry. When I try to use the the PreparedGeometry after dropping the original one I get various errors and sigabrts, including messages like `pure virtual method called` (which could have been a coincidence).

GuillaumeGomez:
Fixes #86.


GuillaumeGomez:
@gauteh: I'll check if there are other cases like this one too.

gauteh:
Thanks a lot!

PR #99, opened 2022-03-21
add static feature for building geos-sys with statically linked libgeos

gauteh:
Depends on https://github.com/georust/geos-sys/pull/21


GuillaumeGomez:
I cherry-picked your commit into https://github.com/georust/geos/pull/100. I'll make releases very soon.

GuillaumeGomez:
New versions have been published.

gauteh:
Thanks. I think that the `exclude` here: https://github.com/georust/geos/blob/master/sys/geos-src/Cargo.toml#L11 prevents the source from being checked out. You could potentially skip other large parts of the source tree if it is a problem: https://github.com/alexcrichton/openssl-src-rs/blob/main/Cargo.toml#L11

PR #104, opened 2022-03-22
geos-src: don't exclude source from package

gauteh:


GuillaumeGomez:
It doesn't work though because the size of `source` is too big. I added after I tried to publish the crate.

GuillaumeGomez:
Do you have the error by any chance? Normally, it should just work since cargo is supposed to checkout submodules as well. Or am I missing something?

GuillaumeGomez:
Would `exclude = ["source/*"]` work better?

gauteh:
No, you can only exclude what is not needed by CMake. See if the commit I just sent makes it small enough. Otherwise we can also exclude `web`, but that requires patching CMakeList.txt since there isn't an option to turn it off.

GuillaumeGomez:
It would be simpler to instead run `git submodule update --init` in `build.rs`.

gauteh:
Yes, that is also possible.

gauteh:
However, the package when I built it now is 8mb which is less than the 10mb limit. Only a couple of mb when compressed.

GuillaumeGomez:
And if `libgeos` makes an update over-reaching this limit, we'll be back to this issue. Wouldn't it be better to just have to update the submodule and not worry about it being too big directly?

gauteh:
Sure. There are other crates that do it that way.

GuillaumeGomez:
I let you handle the update? ;)

GuillaumeGomez:
I think the current implementation is fine but I can wait for you to test more things. Please tell me once it's ready for you so I can review it.

gauteh:
Hi,

I reverted back to ignoring the tests only. We won't get any surprises from libgeos with static since we have to manually update the submodules. I would prefer to deal with it then, presumably by also excluding `web`. I was also made aware that builds happen offline in many cases, e.g. on docs.rs. 

Does that work for you?

Regards, Gaute

GuillaumeGomez:
It's fine for now. Let's hope we won't have bad surprise... I'll make a new release after merging this PR.

gauteh:
Thanks!

GuillaumeGomez:
The three crates have been published.

gauteh:
Thanks, works perfectly.

gauteh:
The current one doesn't work with cargo publish --dry-run. But you could
use the one that excludes the tests.

tir. 22. mar. 2022, 15:55 skrev Guillaume Gomez ***@***.***>:

PR #107, opened 2022-03-23
src: specify predictable libdir

gauteh:


gauteh:
Turns out CMake puts the libs in lib64 sometimes (https://github.com/gauteh/roaring-landmask/runs/5658653361?check_suite_focus=true), this always puts it in lib/.

GuillaumeGomez (approved):

GuillaumeGomez:
Here we go again for a round of release. :)

gauteh:
Hehe, well at least we didn't break the standard no-static version :D

PR #109, opened 2022-04-25
Add Geometry buffer_with_style and buffer_with_params Geos API support.

enerqi:


src/buffer_params.rs
```diff
@@ -0,0 +1,261 @@
...
+use crate::{
+    AsRaw, AsRawMut, ContextHandle, ContextHandling, ContextInteractions, Error, GResult, JoinStyle,
+};
+
+use geos_sys::*;
+use std::sync::Arc;
+
+pub const DEFAULT_MITRE_LIMIT: f64 = 5.0;
+pub const DEFAULT_QUADRANT_SEGMENTS: i32 = 8;
+pub const DEFAULT_SINGLE_SIDED: bool = false;
```
GuillaumeGomez, line 12: That seems like they should all be enums.

enerqi: You would have to explain - I can't see how an enum would fit here. 
The constants are already namespaced by the `buffer_params` module. If those constants became an enum we would then have to create a separate instance of it as the individual cases can't have defaults (a single value) without being a C style enum. Even then a C-style enum is not well typed unless all cases are integers. In usage, such as `buffer_with_style` the cases are not exclusive/disjoint. The C API has separate set param functions (mitre, cap style...) and not a single setParam that can take a union of inputs.

GuillaumeGomez: Because I assume there are other constants like `*_MITRE_*` and for others too.

enerqi: You want the enum to behave like a "constants" namespace instead of just using the buffer params module as the namespace? Seems like the actual constant definitions would then have to be a hashmap mapping from the constant type mitre | quad segments | etc to the constant value.

src/wkb_writer.rs
```diff
@@ -7,7 +7,7 @@ use geos_sys::*;
 use std::convert::TryFrom;
 use std::sync::Arc;
 
-/// The `WKBWriter` type is used to generate `HEX` or `WKB` formatted output from [`Geometry`].
+/// The `WKBWriter` type is used to generate `HEX` or `WKB` formatted output from [`Geometry`](crate::Geometry).
```
GuillaumeGomez, line 10: Why this change?

src/prepared_geometry.rs
```diff
@@ -30,7 +30,7 @@ pub struct PreparedGeometry<'a> {
 }
 
 impl<'a> PreparedGeometry<'a> {
-    /// Creates a new `PreparedGeometry` from a [`Geometry`].
+    /// Creates a new `PreparedGeometry` from a [`Geometry`](crate::Geometry).
```
GuillaumeGomez, line 33: Why this change?

src/wkb_writer.rs
```diff
@@ -202,8 +202,8 @@ impl<'a> WKBWriter<'a> {
         }
     }
 
-    /// Returns the number of dimensions to be used when calling [`WKBWriter::write`]. By default,
-    /// it is 2.
+    /// Returns the number of dimensions to be used when calling [`WKBWriter::write_wkb`].
```
GuillaumeGomez, line 205: Why this change?

src/wkt_writer.rs
```diff
@@ -9,7 +9,7 @@ use geos_sys::*;
 use std::convert::TryFrom;
 use std::sync::Arc;
 
-/// The `WKTWriter` type is used to generate `WKT` formatted output from [`Geometry`].
+/// The `WKTWriter` type is used to generate `WKT` formatted output from [`Geometry`](crate::Geometry).
```
GuillaumeGomez, line 12: Why this change?

src/buffer_params.rs
GuillaumeGomez: Ok, so after taking a look around, I'm not sure if we should keep these constants... If we want to handle default values, might be better to implement `Default` on `BufferParams` instead, no?

src/wkb_writer.rs
enerqi: All these doc string changes are because the links are broken (Geometry is not actually imported in that module), e.g. without the addition when running cargo doc:

> warning: unresolved link to `Geometry`
>   --> src\wkb_writer.rs:10:85
>    |
> 10 | /// The `WKBWriter` type is used to generate `HEX` or `WKB` formatted output from [`Geometry`]
>    |                                                                                     ^^^^^^^^ no item named `Geometry` in scope
>    |
>    = note: `#[warn(rustdoc::broken_intra_doc_links)]` on by default
>    = help: to escape `[` and `]` characters, add '\' before them like `\[` or `\]`

GuillaumeGomez: I see, that's weird it wasn't seen before. Thanks for the fix!

src/buffer_params.rs
enerqi: I can change the `BufferParamsBuilder` to use options instead of default values and only call the C Api to set the values if the option is `Some`...and remove those constants. 
`BufferParams` doesn't really need default as it has `new`. Not setting any options will use the defaults defined in the  C++ API.

GuillaumeGomez: Fine by me!

GuillaumeGomez (approved):

PR #115, opened 2022-08-04
TST: Update tests to pass on GEOS 3.6 - 3.11

brendan-ward:
This updates and simplifies the tests that were breaking on GEOS > 3.6 so that they run on all versions 3.6 - 3.11.

I understand that some of the doctests use geometires from the PostGIS documentation.  However, these are more complex than necessary for simple examples, so I simplified these down for clarity and to ensure they pass.

### DIFF
--- a/src/geometry.rs
+++ b/src/geometry.rs
@@ -89,11 +89,12 @@ pub trait Geom<'a>:
     /// ```
     /// use geos::{Geom, Geometry};
     ///
-    /// let geom = Geometry::new_from_wkt("POLYGON((0 0, 1 1, 1 2, 1 1, 0 0))")
+    /// // Bowtie polygon with self-intersection
+    /// let geom = Geometry::new_from_wkt("POLYGON((0 0, 2 2, 2 0, 0 2, 0 0))")
     ///                     .expect("Invalid geometry");
     /// assert_eq!(
     ///     geom.is_valid_reason(),
-    ///     Ok("Self-intersection[0 0]".to_owned()),
+    ///     Ok("Self-intersection[1 1]".to_owned()),
     /// );
     /// ```
     fn is_valid_reason(&self) -> GResult<String>;
@@ -612,7 +613,8 @@ pub trait Geom<'a>:
     /// let geom = Geometry::create_multipolygon(vec![geom1, geom2])
     ///                     .expect("Failed to build multipolygon");
     ///
-    /// let union_geom = geom.unary_union().expect("unary_union failed");
+    /// let mut union_geom = geom.unary_union().expect("unary_union failed");
+    /// union_geom.normalize().expect("normalize failed");
     ///
     /// assert_eq!(
     ///     union_geom.to_wkt_precision(1).unwrap(),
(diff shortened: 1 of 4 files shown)


src/geometry.rs
```diff
@@ -89,11 +89,12 @@ pub trait Geom<'a>:
     /// ```
     /// use geos::{Geom, Geometry};
     ///
-    /// let geom = Geometry::new_from_wkt("POLYGON((0 0, 1 1, 1 2, 1 1, 0 0))")
+    /// // Bowtie polygon with self-intersection
```
brendan-ward, line 92: The problem with the original is that it shared two edges rather than self-intersecting at a point, so it broke because different versions of GEOS reported different parts of those shared edges as the self-intersection.

This uses the classic "bowtie" self-intersection shape with a single point of self-intersection.

src/geometry.rs
```diff
@@ -627,21 +629,13 @@ pub trait Geom<'a>:
...
-    ///                          POLYGON ((247 50, 77.14285714285714 50, 105 115, 145 195, 178.33333333333334 211.66666666666666, 183.51851851851853 208.7037037037037, 247 50)),\
-    ///                          POLYGON ((-3.75 380, 20.000000000000007 380, 176.66666666666666 223.33333333333334, 178.33333333333334 211.66666666666666, 145 195, 105 235, -3.75 380)),\
-    ///                          POLYGON ((105 115, 105 235, 145 195, 105 115)),\
-    ///                          POLYGON ((20.000000000000007 380, 255 380, 176.66666666666666 223.33333333333334, 20.000000000000007 380)),\
-    ///                          POLYGON ((255 380, 340 380, 340 240, 183.51851851851853 208.7037037037037, 178.33333333333334 211.66666666666666, 176.66666666666666 223.33333333333334, 255 380)),\
-    ///                          POLYGON ((340 240, 340 50, 247 50, 183.51851851851853 208.7037037037037, 340 240)))")
-    ///                             .expect("Invalid geometry");
-    ///
-    /// let mut voronoi = input.voronoi(None::<&Geometry>, 6., false)
+    ///     "GEOMETRYCOLLECTION (POLYGON ((0 0, 0 4, 3 4, 3 0, 0 0)), POLYGON ((6 4, 6 0, 3 0, 3 4, 6 4)))")
```
brendan-ward, line 635: This simplified the example to use far fewer coordinates.  It looks like this:
image

src/geometry.rs
```diff
@@ -2434,22 +2428,15 @@ impl<'a> Geometry<'a> {
     /// ```
     /// use geos::{Geom, Geometry};
     ///
-    /// let geom = Geometry::new_from_wkt("POINT(100 90)").expect("Invalid geometry");
-    /// let small_geom = geom.buffer(25., 8).expect("buffer failed");
-    /// let big_geom = geom.buffer(50., 8).expect("buffer failed");
-    ///
-    /// let union_geom = small_geom.union(&big_geom).expect("union failed");
-    /// let build_area_geom = union_geom.build_area().expect("build_area failed");
+    /// let geom = Geometry::new_from_wkt("GEOMETRYCOLLECTION( \
```
brendan-ward, line 2431: This simplifies the example to use far fewer coordinates, and looks like this:

image

brendan-ward:
Still in progress: all tests passed locally on MacOS but some are still failing on CI / Ubuntu.

brendan-ward:
Tests are failing on CI for unrelated reasons and need to be rerun.

Latest test results here: https://github.com/brendan-ward/geos-rs/runs/7682381950

All tests except memory leaks (existing issue) now passing on all GEOS versions 3.6 - 3.11.

src/voronoi.rs
```diff
@@ -46,45 +51,49 @@ mod test {
...
             POLYGON ((25 295, 25 395, 400 395, 400 369.6542056074766, 221.20588235294116 210.91176470588235, 25 295)))";
 
         let mut expected_output = GGeometry::new_from_wkt(expected_output).unwrap();
 
         expected_output.normalize().unwrap();
         voronoi.normalize().unwrap();
 
-        let same = expected_output.equals(&voronoi).unwrap();
-        assert!(same);
+        assert_eq!(
```
brendan-ward, line 63: Here and next test; due to floating point precision issues there is a tiny amount of difference beyond 10 decimal places.  The fix was to change from equality test to comparing WKT at suitable precision.

src/voronoi.rs
```diff
@@ -158,41 +167,41 @@ mod test {
         let poly = vec![
             Polygon::new(
                 LineString(coords(vec![
+                    (0.5, 0.5),
```
brendan-ward, line 170: Changes here are to manually normalize the ordering within geometry collection and winding order of these polygons to match the output of normalizing the voronoi polygons.

GuillaumeGomez:
Quickly commenting to say I saw the PR. Will review it whenever I can.

GuillaumeGomez:
Looks great, thanks!

PR #116, opened 2022-08-22
Upgrade wkt package version to fix a bug involving multipoint.

aglasscage:
Without the package upgrade we panic with this error:

`thread 'to_geo::test::geom_to_geo_multipoint' panicked at 'called `Result::unwrap()` on an `Err` value: ConversionError("impossible to read wkt: Missing closing parenthesis for type")', src/to_geo.rs:82:59`

### DIFF
--- a/src/to_geo.rs
+++ b/src/to_geo.rs
@@ -1,6 +1,7 @@
 use crate::error::Error;
 use crate::{ConstGeometry, Geom, Geometry as GGeometry};
 use geo_types::Geometry;
+use std::str::FromStr;
 use wkt;
 
 use std::convert::TryFrom;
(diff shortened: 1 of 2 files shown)


src/to_geo.rs
```diff
@@ -18,11 +19,7 @@ impl<'a, 'b$(,$lt)?> TryFrom<&'b $ty_name<'a$(,$lt)?>> for Geometry<f64> {
         let wkt_obj = wkt::Wkt::from_str(&wkt_str)
```
michaelkirk, line 19:
```suggestion
        geo_types::Geometry::try_from_wkt_str(&wkt_str)
```

With wkt 0.10+, you should be able to do this instead, without ever interacting directly with the `Wkt` struct.

michaelkirk: I'm not sure if it's relevant to this crate, but we also added support for converting directly to the inner enum member:
```rust
geo_types::Point::try_from_wkt_str(&wkt_str)
geo_types::LineString::try_from_wkt_str(&wkt_str)
// etc.
```

GuillaumeGomez: Thanks for the suggested improvement! As for the second point, I'm not sure if it's supposed to be handled in here or not...

aglasscage: Thank you! I went ahead and made the suggested change. I also simplified the unit test with more reasonable coordinates as opposed to the exact ones that failed in my system when I noticed the bug.

GuillaumeGomez (approved):

Issue #82, opened 2021-02-04
Upgrade to latest version fo the 'geojson' crate


frewsxcv:
We're a version behind right now

PR #118, opened 2023-01-30
Update geojson version to 0.24

GuillaumeGomez:
Fixes https://github.com/georust/geos/issues/82.


lnicola:
This is a breaking change, right?

GuillaumeGomez:
Yes. I'll update the crate version accordingly.

PR #119, opened 2023-01-30
Fix clippy warnings

GuillaumeGomez:


lnicola:
Might be worth adding a `rust-version` (this requires 1.58, perhaps newer) clause now.

lnicola:
Clippy checks, but doesn't fix build scripts :smile:.

GuillaumeGomez:
> Might be worth adding a `rust-version` (this requires 1.58, perhaps newer) clause now.

Yes absolutely!

Issue #117, opened 2023-01-29
Expose `copyFromBuffer`, `copyToBuffer`, `copyFromArrays`, `copyToArrays`


kylebarron:
👋 Hello!

GEOS 3.10+ includes new APIs to interact between a `CoordSeq` and external buffers/arrays. This would be useful for me because I'm trying to make interop between [GeoArrow](https://github.com/geoarrow/geoarrow) and GEOS faster. I'm guessing that these should be faster than a manual approach with iterating because `copyFromBuffer` should be essentially just a `memcpy`.

Would you take a PR for this?

PR #121, opened 2023-02-07
Implement `copyFromBuffer`, `copyFromArrays`, `copyToBuffer` and `copyToArrays`

kylebarron:
I'm very inexperienced at C/rust ffi, so happy to get some feedback before adding all four functions and tests.

Closes https://github.com/georust/geos/issues/117

### DIFF
--- a/src/coord_seq.rs
+++ b/src/coord_seq.rs
@@ -8,6 +8,9 @@ use geos_sys::*;
 use std::convert::TryFrom;
 use std::sync::Arc;
 
+#[cfg(feature = "v3_10_0")]
+type AsArrayOutput = (Vec<f64>, Vec<f64>, Option<Vec<f64>>, Option<Vec<f64>>);
+
 /// `CoordSeq` represents a list of coordinates inside a [`Geometry`].
 ///
 /// # Example
@@ -58,12 +61,10 @@ impl<'a> CoordSeq<'a> {
     /// assert_eq!(coord_seq2.get_x(1), Ok(1.));
     /// ```
     pub fn new(size: u32, dims: CoordDimensions) -> GResult<CoordSeq<'a>> {
-        match ContextHandle::init_e(Some("CoordSeq::new")) {
-            Ok(context_handle) => unsafe {
-                let ptr = GEOSCoordSeq_create_r(context_handle.as_raw(), size, dims.into());
-                CoordSeq::new_from_raw(ptr, Arc::new(context_handle), size, dims.into(), "new")
-            },
-            Err(e) => Err(e),
+        let context_handle = ContextHandle::init_e(Some("CoordSeq::new"))?;
+        unsafe {
+            let ptr = GEOSCoordSeq_create_r(context_handle.as_raw(), size, dims.into());
+            CoordSeq::new_from_raw(ptr, Arc::new(context_handle), size, dims.into(), "new")
         }
     }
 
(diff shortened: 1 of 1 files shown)


src/coord_seq.rs
```diff
@@ -149,6 +149,93 @@ impl<'a> CoordSeq<'a> {
...
 
+    #[cfg(any(feature = "v3_10_0", feature = "v3_11_0"))]
+    pub fn new_from_buffer<T: AsRef<[f64]>>(
+        data: &[f64],
+        size: u32,
+        dims: u32,
+        has_z: bool,
+        has_m: bool,
+    ) -> GResult<CoordSeq<'a>> {
+        let data_ptr = &data[0] as *const f64;
```
frewsxcv, line 160:
```suggestion
        let data_ptr = data.as_ptr();
```

https://doc.rust-lang.org/std/primitive.slice.html#method.as_ptr

kylebarron: I knew there had to be a simpler way 😂

src/coord_seq.rs
```diff
@@ -149,6 +149,93 @@ impl<'a> CoordSeq<'a> {
         }
     }
 
+    #[cfg(any(feature = "v3_10_0", feature = "v3_11_0"))]
```
GuillaumeGomez, line 152:
```suggestion
    #[cfg(feature = "v3_10_0")]
```

Any "higher" feature enables the "lower" ones recursively.

src/coord_seq.rs
```diff
@@ -149,6 +149,93 @@ impl<'a> CoordSeq<'a> {
...
+                CoordSeq::new_from_raw(
+                    ptr,
+                    Arc::new(context_handle),
+                    size,
+                    dims.into(),
+                    "new_from_buffer",
+                )
+            },
+            Err(e) => Err(e),
+        }
```
GuillaumeGomez, line 185: I think this will work too:

```suggestion
        let context_handle = ContextHandle::init_e(Some("CoordSeq::new_from_buffer"))?;
        let ptr = GEOSCoordSeq_copyFromBuffer_r(
                context_handle.as_raw(),
                data_ptr,
                size as _,
                has_z as _,
                has_m as _,
            );
            CoordSeq::new_from_raw(
                ptr,
                Arc::new(context_handle),
                size,
                dims.into(),
                "new_from_buffer",
            )
```

kylebarron: Makes sense; I was copying from `new()` which used match

GuillaumeGomez: Some parts of this crate are **old**. So nothing surprising. ;)

src/coord_seq.rs
```diff
@@ -149,6 +147,113 @@ impl<'a> CoordSeq<'a> {
...
+        };
+        let m_ptr = if let Some(m) = m {
+            assert_eq!(x.len(), m.len(), "Arrays have different lengths.");
+            dims += 1;
+            m.as_ptr()
+        } else {
+            std::ptr::null()
+        };
+
+        let size = x.len() as u32;
```
GuillaumeGomez, line 235: No need to cast here. Or to store the size, you can directly put it below (I'll comment directly on it).

src/coord_seq.rs
```diff
@@ -149,6 +147,113 @@ impl<'a> CoordSeq<'a> {
...
+
+        let context_handle = ContextHandle::init_e(Some("CoordSeq::new_from_arrays"))?;
+        unsafe {
+            let ptr = GEOSCoordSeq_copyFromArrays_r(
+                context_handle.as_raw(),
+                x.as_ptr(),
+                y.as_ptr(),
+                z_ptr,
+                m_ptr,
+                size as _,
```
GuillaumeGomez, line 245: So as mentioned above, you can directly do:

```suggestion
                x.len() as _,
```

kylebarron:
Is this crate concerned at all with `m` values? If they aren't supported anywhere else, should I remove them from these functions?

kylebarron:
I added a couple more doctests. I think the `new_from_buffer` function signature should be simplified still, especially because it currently lets you pass in `size: 3` but `has_z: false` and `has_m: false`. If we assume we never have m values then that makes this a little simpler.

The other thing to note is in `get_as_buffer`/`copyToBuffer`. It appears that GEOS always stores coordinates internally as 3 dimensions, so `copyToBuffer` will fill in `NaN` values for z coords. Should the API here handle repacking into a 2d buffer? For my case that's unnecessary because I'll be repacking again into arrow memory anyways

src/coord_seq.rs
```diff
@@ -383,6 +486,110 @@ impl<'a> CoordSeq<'a> {
...
+        unsafe {
+            GEOSCoordSeq_copyToArrays_r(
+                self.get_raw_context(),
+                self.as_raw(),
+                x.as_mut_ptr(),
+                y.as_mut_ptr(),
+                z.as_mut()
+                    .map(|arr| arr.as_mut_ptr())
+                    .unwrap_or_else(|| std::ptr::null_mut()),
+                std::ptr::null_mut(),
```
GuillaumeGomez, line 586: We allow to set `m` but we don't get it back? Seems a bit strange no?

src/coord_seq.rs
```diff
@@ -383,6 +486,110 @@ impl<'a> CoordSeq<'a> {
...
+    ///
+    /// let output_arrays = coords.get_as_arrays()
+    ///                           .expect("failed to get arrays");
+    ///
+    /// assert_eq!(output_arrays.0, vec![0., 3., 6.]);
+    /// assert_eq!(output_arrays.1, vec![1., 4., 7.]);
+    /// assert_eq!(output_arrays.2, Some(vec![100., 200., 300.]));
+    /// ```
+    #[cfg(feature = "v3_10_0")]
+    pub fn get_as_arrays(&self) -> GResult<(Vec<f64>, Vec<f64>, Option<Vec<f64>>)> {
```
GuillaumeGomez, line 567: Method name should be `as_arrays`.

src/coord_seq.rs
```diff
@@ -383,6 +486,110 @@ impl<'a> CoordSeq<'a> {
...
+    ///
+    /// let output_buffer = coords.get_as_buffer()
+    ///                           .expect("failed to get buffer");
+    ///
+    /// let expected_output_buffer = vec![0., 1., f64::NAN, 2., 3., f64::NAN, 4., 5., f64::NAN];
+    /// assert_eq!(output_buffer[0], expected_output_buffer[0]);
+    /// assert_eq!(output_buffer[3], expected_output_buffer[3]);
+    /// ```
+    #[cfg(feature = "v3_10_0")]
+    pub fn get_as_buffer(&self) -> GResult<Vec<f64>> {
```
GuillaumeGomez, line 511: Method name should be `as_buffer`.

GuillaumeGomez:
> I added a couple more doctests. I think the `new_from_buffer` function signature should be simplified still, especially because it currently lets you pass in `size: 3` but `has_z: false` and `has_m: false`. If we assume we never have m values then that makes this a little simpler.

`size` is independent from the `m` and everything as far as I can see: https://github.com/libgeos/geos/blob/0a55b3af552470754893485b51407cc0219dbaae/capi/geos_ts_c.cpp#L2435-L2443

kylebarron:
> `size` is independent from the `m` as far as I can see: [libgeos/geos@`0a55b3a`/capi/geos_ts_c.cpp#L2435-L2443](https://github.com/libgeos/geos/blob/0a55b3af552470754893485b51407cc0219dbaae/capi/geos_ts_c.cpp#L2435-L2443)

That's so interesting; I should've known to look at the source! So passing in a buffer of `XY` is not a straight memcpy

src/coord_seq.rs
```diff
@@ -149,6 +147,111 @@ impl<'a> CoordSeq<'a> {
...
+        y: &[f64],
+        z: Option<&[f64]>,
+        m: Option<&[f64]>,
+    ) -> GResult<CoordSeq<'a>> {
+        assert_eq!(x.len(), y.len(), "Arrays have different lengths.");
+
+        let mut dims: u32 = 2;
+        let z_ptr = if let Some(z) = z {
+            assert_eq!(x.len(), z.len(), "Arrays have different lengths.");
+            dims += 1;
```
GuillaumeGomez, line 222: Are you sure we're supposed to compute the dimensions ourselves btw?

kylebarron: I'm not sure I fully understand your question... `CoordSeq::new_from_raw` requires the number of dimensions as an argument. We could make the user pass it in, but we can also infer it based on whether `z` or `m` is provided?

GuillaumeGomez: You're right, my bad.

kylebarron:
> > I added a couple more doctests. I think the `new_from_buffer` function signature should be simplified still, especially because it currently lets you pass in `size: 3` but `has_z: false` and `has_m: false`. If we assume we never have m values then that makes this a little simpler.
> 
> `size` is independent from the `m` as far as I can see: [libgeos/geos@`0a55b3a`/capi/geos_ts_c.cpp#L2435-L2443](https://github.com/libgeos/geos/blob/0a55b3af552470754893485b51407cc0219dbaae/capi/geos_ts_c.cpp#L2435-L2443)

Oops I meant to write `dims`, not `size`. `size` is the number of coordinates in the array; `dims` is the number of dimensions; we should be able to infer `dims` from `has_z` and `has_m`

GuillaumeGomez:
Time for the last part of the PR: fixing CI errors. :)

Don't worry about errors from valgrind though. Also, don't forget to rebase on `master`.

kylebarron:
I updated with better docstrings, a bug fixed in `as_buffer`, better handling of m values

GuillaumeGomez:
Clippy is still unhappy.

GuillaumeGomez:
Clippy is still failing for the same reason.

kylebarron:
> Clippy is still failing for the same reason.

It is? The Clippy output is different in my last two commits. The first time it had a warning [related to my changes](https://github.com/georust/geos/actions/runs/4158248730/jobs/7194256509#step:5:36); now all the clippy warnings are unrelated to my changes. I can change the lifetime syntax there if you'd like

GuillaumeGomez:
Oh sorry, didn't pay attention. I opened https://github.com/georust/geos/pull/122 to fix it. In the meantime, please squash your commits.

kylebarron:
I guess this is why on my own repos I usually let github squash my commits for me, because I'm inept at git 😂

kylebarron:
I guess there's no way for me to re-open this PR, even if I push a new commit to the branch now; sorry about that! I'll have to open a new PR

kylebarron:
My issue was I followed the second suggestion in https://stackoverflow.com/a/25357146 and ran
```
git reset --soft $(git merge-base master HEAD)
```
but then ran `git push -f`, forgetting to commit again 😆

GuillaumeGomez (approved):

GuillaumeGomez:
> thanks! didn't know you could do that

Me neither until we ended up in this situation. ;)

GuillaumeGomez:
Thanks!

kylebarron:
thanks for your help!

PR #124, opened 2023-03-29
Fix 'from_geo' example in README

mthh:
This fixes the 'from_geo' example in the README (and so in the doc generated by `cargo doc`, as it includes the README) by removing the `use geos::from_geo::TryInto;` statement (otherwise cargo complains that this trait is private).

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,7 +53,6 @@ to use all geos algorithms.
 Complete example can be found in `examples/from_geo.rs`
 
 ```rust,ignore
-use geos::from_geo::TryInto;
 use geos::geo_types::{LineString, Coordinate, Polygon};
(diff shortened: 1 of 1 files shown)


GuillaumeGomez (approved):

GuillaumeGomez:
Considering it's `ignore`, it shouldn't be compiled so I'm a bit surprised. But thanks for the fix in any case!

mthh:
Indeed, but a user might want to copy/paste the example from the README / from the doc (and thus come across this error - that's how I found it).

GuillaumeGomez:
Absolutely, hence why I'm approving this PR. Thanks again!

PR #136, opened 2023-06-12
Add-descriptive-comment-to-comput_vornoi-function-tolerance-parameter

MitchellNeill:
Adds a comment describing the tolerance parameter of the compute_voronoi function. This is useful as the parameter is described as finicky by the underlying c library. Having this comment would have saved me a headache.

Open to suggestions on better ways to expose this information.

### DIFF
--- a/src/voronoi.rs
+++ b/src/voronoi.rs
@@ -6,6 +6,10 @@ use std::borrow::Borrow;
 use std::convert::TryInto;
 
 /// Available using the `geo` feature.
+/// About the `tolerance` argument, the underlying C library mentions:
+/// > snapping tolerance to use for improved robustness. A tolerance of 0.0 specifies that no snapping
+/// > will take place. This argument can be finicky and is known to cause the algorithm to fail in
+/// > several cases. If you're using tolerance and getting a failure, try setting it to 0.0.
 pub fn compute_voronoi<T: Borrow<Point<f64>>>(
     points: &[T],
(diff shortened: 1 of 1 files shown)


src/voronoi.rs
```diff
@@ -9,6 +9,10 @@ use std::convert::TryInto;
 pub fn compute_voronoi<T: Borrow<Point<f64>>>(
     points: &[T],
     envelope: Option<&GGeometry>,
+    /// From the underlying c library
+    /// "snapping tolerance to use for improved robustness. A tolerance of 0.0 specifies that no snapping
+    /// will take place. This argument can be finicky and is known to cause the algorithm to fail in
+    /// several cases. If you're using tolerance and getting a failure, try setting it to 0.0."
```
GuillaumeGomez, line 15: Doc comments are not supported on function arguments, so you'll need to put it on the function. Also, we could use "quote" syntax instead:

```
    /// From the underlying C library:
    /// > snapping tolerance to use for improved robustness. A tolerance of 0.0 specifies that no snapping
    /// > will take place. This argument can be finicky and is known to cause the algorithm to fail in
    /// > several cases. If you're using tolerance and getting a failure, try setting it to 0.0.
```

MitchellNeill: I can do that, how does the new version look?

GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

Issue #137, opened 2023-09-30
`get_coord_seq` without extra clone


kylebarron:
From the docstring of `get_coord_seq`:

> Note: this clones the underlying CoordSeq to avoid double free (because CoordSeq handles the object ptr and the CoordSeq is still owned by the geos geometry) if this method’s performance becomes a bottleneck, feel free to open an issue, we could skip this clone with cleaner code.

My use case is to bind GEOS algorithms to the GeoArrow memory layout (an efficient geometry layout for arrays of geometries, see [geoarrow.org](https://geoarrow.org/) and my WIP rust implementation at https://github.com/geoarrow/geoarrow-rs). My current plan is to always store geometries before and after each operation in GeoArrow memory, and therefore GEOS objects are totally ephemeral during an operation. So the process goes like

1. Take an array of e.g. polygons
2. Iterate over the array, converting each into GEOS objects
3. Apply a GEOS operation, say, buffering
4. Construct a new GeoArrow array from the polygon outputs

Therefore the IO to and from GEOS objects is really important to me, because it's overhead for every operation on the array.

One option is to improve this `get_coord_seq`, removing a clone. The other possibility for me is to have something like `into_coord_seq` or `into_inner`. Given that I want to consume the GEOS geometry anyways, and only access its coords, this might be easier to implement?

I'd be willing to attempt a PR!

GuillaumeGomez:
Both approaches sound good to me but if the `into_coord_seq` approach is simpler, please go ahead with this one.

Might still be worth to improve `get_coord_seq` in the future though.

kylebarron:
Actually thinking about this a little more.. might one way to solve this be to implement a `ConstCoordSeq` struct just like there's a `ConstGeometry`?

GuillaumeGomez:
That could work too.

PR #138, opened 2023-10-01
WIP ConstCoordSeq

kylebarron:
Closes https://github.com/georust/geos/issues/137

This adds a new struct for `ConstCoordSeq` that contains a const pointer to the GEOS coord seq object.

The ideal API would probably be to have something like the `Geom` trait for coordinate sequences, but I assume that's backwards-incompatible, because users would then also have to import such a trait?

I didn't notice any performance improvement on my relevant benchmark in my geoarrow project, so it's possible I either implemented it wrong or this is such a small performance gain that it's immeasurably small.

### DIFF
--- a/src/coord_seq.rs
+++ b/src/coord_seq.rs
@@ -30,6 +30,16 @@ pub struct CoordSeq<'a> {
     nb_lines: usize,
 }
 
+/// Representation of an immutable GEOS coordinate sequence. Since it's only a view over another
+/// GEOS geometry's data, only immutable operations are implemented on it.
+pub struct ConstCoordSeq<'a, 'b> {
+    pub(crate) ptr: PtrWrap<*const GEOSCoordSequence>,
+    pub(crate) original: &'b Geometry<'a>,
+    // pub(crate) context: ContextHandle<'a>,
+    nb_dimensions: usize,
+    nb_lines: usize,
+}
+
 impl<'a> CoordSeq<'a> {
     /// Creates a new `CoordSeq`.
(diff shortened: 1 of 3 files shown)


src/coord_seq.rs
```diff
@@ -928,3 +937,475 @@ impl<'a> ContextHandling for CoordSeq<'a> {
...
+    ) -> GResult<ConstCoordSeq<'a, 'b>> {
+        if ptr.is_null() {
+            let extra = if let Some(x) = original.context.get_last_error() {
+                format!("\nLast error: {x}")
+            } else {
+                String::new()
+            };
+            return Err(Error::NoConstructionFromNullPtr(format!(
+                "CoordSeq::{caller}{extra}",
+            )));
```
GuillaumeGomez, line 957: You're calling two `format!` here, which isn't great. Could you instead have `format!`s directly into the `let Some(x) else` please?

GuillaumeGomez:
I don't see anything bad in particular in the implementation. So the relevant question here would be: are you sure this clone was this expensive for you? Do you have a flamegraph comparison between before and after?

kylebarron:
> are you sure this clone was this expensive for you

No I'm not 😄 it was an (uneducated?) guess. I need to learn how to better profile rust code 😅

GuillaumeGomez:
Well in any case, this PR seems to be on good track so if you fix the errors, I'll gladly merge it. Take a look at how the other `Const*` types were implemented if you need to find out.

PR #145, opened 2024-02-07
Update geos version to 8.3.1 and geos-sys version to 2.0.6

GuillaumeGomez:


GuillaumeGomez:
Done and sorry again about that.

PR #161, opened 2024-12-02
fix: get_num_geometries failing on empty geom

vladaburian:
Failed call to `GEOSGetNumGeometries_r` is signaled by -1. Zero on the other hand is valid result when geometry is empty.


GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

Issue #160, opened 2024-10-31
Memory Leak


ManeraKai:
I simplified it as much as possible. Here's my Cargo.toml
```toml
[package]
name = "geos-geometry"
version = "0.1.0"
edition = "2021"

[dependencies]
geos = "9.0.0"
rocket = "0.5.1"
```
main.rs:
```rs
#[macro_use]
extern crate rocket;

#[get("/")]
fn index() -> &'static str {
    let _: Vec<geos::Geometry> = (0..2_000_000)
        .map(|_| geos::Geometry::create_empty_point().unwrap())
        .collect();

    "finished"
}

#[launch]
fn rocket() -> _ {
    rocket::build().mount("/", routes![index])
}
```
GEOS version: `3.12.1-CAPI-1.18.1`

Every time I request `curl http://127.0.0.1:8000/`, I get a +3GB spike in memory. It never frees it. It continues allocating more and more until my OS runs out of memory and crashes. I'm on Ubuntu 24.04.

If you think this is related to rocket, I don't think it is. Here's a code where I only used `Vec<i64>`:
```rs
#[macro_use]
extern crate rocket;

#[get("/")]
fn index() -> &'static str {
    let n: i64 = 1_000_000_000;
    let mut array: Vec<i64> = Vec::with_capacity(n as usize);
    for i in 0..n {
        array.push(i);
    }
    "finished"
}

#[launch]
fn rocket() -> _ {
    rocket::build().mount("/", routes![index])
}
```
This code spikes memory to about 7GB, but it instantly frees after the request is completed.

Here's the geos-sys version:
```rs
use geos::sys::{
    GEOSGeom_createEmptyPoint_r, GEOSGeom_destroy_r, GEOSGeom_t, GEOS_finish_r, GEOS_init_r,
};

#[macro_use]
extern crate rocket;

#[get("/")]
fn index() -> &'static str {
    let ctx = unsafe { GEOS_init_r() };

    let n = 10_000_000;
    let mut geometries: Vec<*mut GEOSGeom_t> = Vec::with_capacity(n);

    for _ in 0..n {
        let geom;
        unsafe {
            geom = GEOSGeom_createEmptyPoint_r(ctx);
        }
        geometries.push(geom);
    }

    for geom in geometries {
        unsafe {
            GEOSGeom_destroy_r(ctx, geom);
        }
    }

    unsafe {
        GEOS_finish_r(ctx);
    }

    "finished"
}

#[launch]
fn rocket() -> _ {
[...]

Issue #141, opened 2023-11-15
Thread safety of geos context and lifetimes


vladaburian:
Hi,

I have few questions/suggestions about geos context and lifetimes. (I'm literally couple of days into learning rust so I may not see some obvious things.)

### 1) GEOS context

I don't understand why each geometry keeps geos context and why is this context cloned over to other derived geometries (even though the cloned rust context is referencing the same instance of geos context).

Geos context is ment to be used as a thread safe mechanism to propagate C++ exceptions through C api and to keep some temporary/helper data AFAIK (like buffer to format error messages or geos Point instance to be re-used as a temporary object). But rust binding does allow non-thread safe usage.

I think it would be good to have thread local instances of geos context and to use those temporarily just to call C geos api. This wayite would also prevent unnecessary allocations of new contexts in many operations (like creating new non-derived geometries, new coordinate sequences, etc.).  In the end the whole context mechanism would be private (and transparent to the users of rust api).

The only thing in geos context which is not primarily related to thread safety is configuration of WKB dims and byte order. But I feel like it does not belong there and also it has been deprecated.

There is draft of what I mean: https://github.com/georust/geos/commit/8c096818dd8b73dfdee93ccad77dcec743c48e0f

links:
- c geos context structure: https://github.com/libgeos/geos/blob/3.12/capi/geos_ts_c.cpp#L202-L214
- using context to store exceptions: https://github.com/libgeos/geos/blob/3.12/capi/geos_ts_c.cpp#L415-L435
- deprecation of wkb setters/getters on context: https://github.com/libgeos/geos/blob/3.12/capi/geos_c.h.in#L5870-L5908
- re-use of Point from context: https://github.com/libgeos/geos/blob/3.12/capi/geos_ts_c.cpp#L3500-L3507

### 2) Lifetimes

I don't understand why Geometry struct (and other structs) has explicit lifetime and why all functions returning new derived geometries also have the same lifetime. Because derived geometries resulting from geometric operations, etc. are just new and owned geometries non-related to the input geometry/geometries.

(It's true that derived geos geometries do pass around non-owned pointer to geometry factory, which is also part of geos context, but which cannot be changed with C api and which is always set to the single global instance.)

It seems to me that all lifetime parameters could be removed and the code would still be correct. And then it would be possible to pass Geometry freely around without any constraints. (But as I wrote, I do not have much experience with rust yet. Maybe I'm missing something.)

[...]

PR #164, opened 2024-12-03
Tthread-local geos context

vladaburian:
This PR is follow-up of my previous issue #141 and also reaction to #160. (I don't think there is a memory leak, but this PR does lower memory footprint of every geometry by little more than 1kB as that is roughly the size of geos context, which would be reused with my PR and not allocated for every geometry.)

I see geos context as just an implementation detail of propagating error messages from c++ geos exceptions through geos c api in reentrant way. To address questions in #141: I dont't think the context is useful from user's point of view and I would say it is better/simpler to not have it in public api at all. I also think that context should not be a property of geometry for the same reason (and because unnecessarily used memory). And also I don't see it useful to be able to set own error handler as the context is already used to propagate geos errors to `GResult`. The original c++ geos api has nothing like the context and also python geos wrapper `shapely` is using geos context only as an internal implementation detail and is not accessible to users.

Another reason for this PR is that geos context is not thread-safe (it's not `Sync`), so this PR makes things correct. Although right now it seems that the worst case is possibly corrupted error message which is not severe problem really, but still.

So this is my suggestion :)

### DIFF
--- a/src/functions.rs
+++ b/src/functions.rs
@@ -1,14 +1,13 @@
-use crate::context_handle::PtrWrap;
+use crate::context_handle::{with_context, PtrWrap};
 use crate::enums::*;
 use crate::error::{Error, GResult, PredicateType};
 use crate::geometry::Geometry;
-use crate::{AsRawMut, ContextHandle, ContextHandling, Geom};
+use crate::{AsRawMut, ContextHandle, Geom};
 use geos_sys::*;
 use std::convert::TryFrom;
 use std::ffi::CStr;
 use std::os::raw::c_char;
 use std::str;
-use std::sync::Arc;
 
 // We need to cleanup only the char* from geos, the const char* are not to be freed.
 // this has to be checked method by method in geos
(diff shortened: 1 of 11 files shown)


vladaburian:
To the actual code change: instead of allocating new geos contexts I'm just reusing the same thread-local contexts via `with_context` function. The context is used only to make c api call and is not passed around or shared in any other way.

src/functions.rs
```diff
@@ -159,36 +138,33 @@ pub fn segment_intersection(
...
-            }
-        },
-        Err(e) => Err(e),
-    }
+    with_context(|ctx| unsafe {
+        let mut cx = 0.;
+        let mut cy = 0.;
+
+        let ret = GEOSSegmentIntersection_r(
+            ctx.as_raw(),
```
GuillaumeGomez, line 146: Is it needed here? Couldn't `with_context` be called inside `GEOSSegmentIntersection_r` methods directly?

GuillaumeGomez:
This is super cool! Just one question and then seems ready for me to go.

src/functions.rs
lnicola: `GEOSSegmentIntersection_r` is a GEOS C API, how would that work?

GuillaumeGomez: Ah it's a function. My bad, then that answers my question.

GuillaumeGomez (approved):

GuillaumeGomez:
Again, great improvement, thanks!

vladaburian:
Thank you :)

PR #175, opened 2024-12-03
fix: lifetime of geometry referenced by PreparedGeometry (version 1)

vladaburian:


src/prepared_geometry.rs
```diff
@@ -21,11 +23,13 @@ use geos_sys::*;
 ///
 /// assert_eq!(prepared_geom.contains(&geom2), Ok(true));
 /// ```
-pub struct PreparedGeometry {
+pub struct PreparedGeometry<'a> {
     ptr: PtrWrap<*const GEOSPreparedGeometry>,
+    geom: Option<Geometry>,
```
GuillaumeGomez, line 28: What is this field used for?

src/prepared_geometry.rs
```diff
@@ -37,18 +41,28 @@ impl PreparedGeometry {
...
     /// ```
-    pub fn new<G: Geom>(g: &G) -> GResult<PreparedGeometry> {
+    pub fn new<G: Geom>(g: &'a G) -> GResult<PreparedGeometry<'a>> {
         with_context(|ctx| unsafe {
             let ptr = GEOSPrepare_r(ctx.as_raw(), g.as_raw());
             PreparedGeometry::new_from_raw(ptr, ctx, "new")
         })
     }
 
+    pub fn new_owning(g: Geometry) -> GResult<PreparedGeometry<'static>> {
```
GuillaumeGomez, line 51: Not a big fan of this but I suppose it's for the case where someone wants to return both a `Geometry` and a `PreparedGeometry` right? If so I don't think it's a case that we should allow. There are many ways to handle it and I don't think returning both types from a function is a good Rust use case.

vladaburian: Yes, it's for case when someone wants to be able to move `PreparedGeometry` freely around, encapsulate in inside some struct etc. I think it's valid case, not something that should be difficult or impossible to do. I have posted some examples in the issue. I don't see the other ways to handle it (or perhaps the answer is to not encapsulate `PreparedGeometry` anywhere, but that sound very limiting.)

src/prepared_geometry.rs
vladaburian: That's so that `PreparedGeometry` can optionally own the related `Geometry` (and have optionally `static` lifetime). It either has `static` lifetime and owns the geometry, or has limited lifetime tied to external `Geometry` (in which case the `geom` field is `None`).

GuillaumeGomez:
Done in https://github.com/georust/geos/pull/179.

PR #176, opened 2024-12-03
fix: lifetime of geometry referenced by PreparedGeometry (version 2)

vladaburian:


GuillaumeGomez:
I prefer the approach in #175. :)

GuillaumeGomez:
Done in https://github.com/georust/geos/pull/179.

PR #182, opened 2025-03-25
Add bindings to GEOSGeoJSONWriter and `new_from_geojson`

Oreilles:
This PR adds a new struct `GeoJSONWriter` with same properties and implementation as `WKTWriter`.  Just like GEOS C++ API, writing is done with the two methods:
```rs
pub fn write<G: Geom>(&mut self, geometry: &G) -> GResult<String>;
pub fn write_formatted<G: Geom>(&mut self, geometry: &G, indent: i32) -> GResult<String>;
```
The writer can be called with the two added functions in the Geometry trait:
```rs
fn to_geojson(&self) -> GResult<String>;
fn to_geojson_formatted(&self, indent: i32) -> GResult<String>;
```
Finally, the function `new_from_geojson` is also added to the Geometry trait to bring parity with `new_from_wkt`.

### DIFF
--- a/src/geojson_writer.rs
+++ b/src/geojson_writer.rs
@@ -1,0 +1,113 @@
+use crate::context_handle::{with_context, PtrWrap};
+use crate::error::Error;
+use crate::functions::*;
+use crate::{AsRaw, AsRawMut, ContextHandle, GResult, Geom};
+use geos_sys::*;
+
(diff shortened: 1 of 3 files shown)


GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

PR #183, opened 2025-03-25
Fix `Precision` enum values

Oreilles:
Fix `Precision` enum values.


GuillaumeGomez (approved):
Outch, nice catch. Thanks!

Issue #156, opened 2024-08-31
`get_srid` and `set_srid` should return/accept `i32` instead of `usize`


Oreilles:
GEOS reads the srid from WKB as a signed integer, and also stores it as a signed integer. This behavior should probably be preserved in this library, instead of trying to cast it to usize (and returning `Err` if not possible).

PR #184, opened 2025-03-25
Change SRID type from usize to i32

Oreilles:
This makes it in line with GEOS. This PR also contains a fix for geometries having -1 as an SRID, which is not an error code for `GEOSGetSRID_r` (0 is, but it's better to keep it that way, as it already implicetly means "no SRID").

Fixes https://github.com/georust/geos/issues/156

### DIFF
--- a/src/geometry.rs
+++ b/src/geometry.rs
@@ -2010,11 +2010,7 @@ impl$(<$lt>)? Geom for $ty_name$(<$lt>)? {
     fn get_srid(&self) -> GResult<i32> {
         with_context(|ctx| unsafe {
             let ret = GEOSGetSRID_r(ctx.as_raw(), self.as_raw());
-            if ret < 1 {
-                Err(Error::GenericError("GEOSGetSRID_r failed".to_owned()))
-            } else {
-                Ok(ret as _)
-            }
+            Ok(ret as _)
         })
     }
 
--- a/src/geometry.rs
+++ b/src/geometry.rs
@@ -1091,7 +1091,7 @@ pub trait Geom: AsRaw<RawType = GEOSGeometry> {
     /// point_geom.set_srid(4326);
     /// assert_eq!(point_geom.get_srid(), Ok(4326));
     /// ```
-    fn get_srid(&self) -> GResult<i32>;
+    fn get_srid(&self) -> GResult<libc::c_int>;
     /// Returns the precision of `self`.
     ///
     /// Available using the `v3_6_0` feature.
@@ -2007,7 +2007,7 @@ impl$(<$lt>)? Geom for $ty_name$(<$lt>)? {
         })
     }
 
-    fn get_srid(&self) -> GResult<i32> {
+    fn get_srid(&self) -> GResult<libc::c_int> {
         with_context(|ctx| unsafe {
             let ret = GEOSGetSRID_r(ctx.as_raw(), self.as_raw());
             Ok(ret as _)
@@ -2641,8 +2641,8 @@ impl Geometry {
     /// point_geom.set_srid(4326);
     /// assert_eq!(point_geom.get_srid(), Ok(4326));
     /// ```
-    pub fn set_srid(&mut self, srid: i32) {
-        with_context(|ctx| unsafe { GEOSSetSRID_r(ctx.as_raw(), self.as_raw_mut(), srid as _) })
+    pub fn set_srid(&mut self, srid: libc::c_int) {
+        with_context(|ctx| unsafe { GEOSSetSRID_r(ctx.as_raw(), self.as_raw_mut(), srid) })
     }
 
     /// Normalizes `self` in its normalized/canonical form. May reorder vertices in polygon rings,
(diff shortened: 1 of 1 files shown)


src/geometry.rs
```diff
@@ -1091,7 +1091,7 @@ pub trait Geom: AsRaw<RawType = GEOSGeometry> {
     /// point_geom.set_srid(4326);
     /// assert_eq!(point_geom.get_srid(), Ok(4326));
     /// ```
-    fn get_srid(&self) -> GResult<usize>;
+    fn get_srid(&self) -> GResult<i32>;
```
GuillaumeGomez, line 1094: Shouldn't it be `isize` instead?

Oreilles: I don't think so, in GEOS it is `int` which I believe is always 32 bit (see [here] (https://github.com/libgeos/geos/blob/main/capi/geos_c.cpp#L1042-L1052)). The WKB parser is more explicit about it, where the SRID is read as `int32_t` (https://github.com/libgeos/geos/blob/main/include/geos/io/ByteOrderDataInStream.h#L70-L78).

GuillaumeGomez: Trying to guess the size of a C type is **never** a good idea. Unless it uses a specific size, never assume it.

Oreilles: True, but in this case most popular implementation of the WKB standard (GEOS, PostGIS, Spatialite) are using `int32_t` as the input and output type for SRIDs so making it 64 bits here doesn't seem like the appropriate decision. If you're reading data from elswhere, then you only need 32 bits. If you're writing data, then allowing 64 bits here will potentially lead to unexpected behaviors further down.

GuillaumeGomez: Currently yes, but who knows in the future? If we use `isize`, we don't have to worry about it anymore. With `i32`, we might have some annoying bugs in the future (even if it's in a long time).

Oreilles: If we use `isize` for SRIDs, we'll have issues *right now*, because that isn't compatible with how WKB stores it, and that's not going to be compatible with any platform that uses 32 bits or less for `int`, which is the vast majority (if not all) of them including 64 bits ones.

GuillaumeGomez: I don't see where the incompatibility is. It's rust code, it only interacts with C code through the API we provide.

Oreilles: There are two incompatibilities, the first one being that GEOS API allows writing EWKB, and you cannot write 64 bits in the SRID field of EWKB. The second one is that for the vast majority (if not all) implementations of the stdlib, `int` is 32 bits (*maybe* 16 bits in some rare cases). Which means we're almost guaranteed overflow when writing our *isize* to the GEOS C `int` representation.

GuillaumeGomez: Hence why we have `TryFrom` trait.

Oreilles: It's also a limitation in the other direction: if this were to be compiled on a platform where `isize` was 16 bits, then reading valid EWKB with high SRID values would fail.

Oreilles: In any case, I don't understand why we'd want to allow 64 bits for SRIDs anyway considering it's almost a guarantee that it's going to fail in GEOS (do you know any context where a C `int` is 64 bits ?), it's always going to fail in the only IO format that supports it, and every other implementation that I know of (be it that of PostgreSQL, Spatialite, MySQL, Oracle, SQL Server) uses 32 bits ints specifically because they either rely on EWKB as their internal storage format or as their primary IO format. SRIDs are at most 900913 in related databases (PostGIS or PROJ) so it's not like we'll run out of values anytime soon.

michaelkirk: (just a drive by comment from someone who is not a `geos` maintainer)

As @Oreilles said, for anyone on 64bit (most people at this point) using `isize` would imply that your range is larger than it is. Sure you can fail at runtime with a Result in that case, but why not have the compiler help you communicate actual requirements to your user at compile time?

rust's `isize` is not the corollary to c's `int`.  If you truly want to match c's type for the rare cases where `int` is 16bits and avoid "guessing the size", you could use libc::c_int. Heres it's impl:

https://docs.rs/libc/latest/src/libc/primitives.rs.html#54
```
cfg_if! {
    if #[cfg(any(target_arch = "avr", target_arch = "msp430"))] {
        pub type c_int = i16;
        pub type c_uint = u16;
    } else {
        pub type c_int = i32;
        pub type c_uint = u32;
    }
}
```

GuillaumeGomez: Then better use `c_int` directly. Like that it kinda solves the whole debate.

Oreilles: Perfect, thanks for the input @michaelkirk.

GuillaumeGomez (approved):

Oreilles:
On an unrelated note, should we do something about those nightly tests ? It's a bit of a downer having that failed CI notification and mark after every PR / commit.

GuillaumeGomez:
Never took time to figure out what's wrong with the linking. But I suppose we can remove them.

PR #188, opened 2025-03-26
Change enums Error type from `&'static str` to `crate::error::Error`

Oreilles:
The rationale is that since they're mostly going to be user among other geos functions, it avoids having to repeatedly  cast their error type manually.


kylebarron:
I'm not a primary maintainer here but approved some of the workflows. I'm not sure why CI is failing.

Oreilles:
CI is failing due to [the depreciation of `actions/cache@v2`](https://github.com/marketplace/actions/cache). Upgrading to version 3 or 4 would allow the tests to run.

GuillaumeGomez (approved):

GuillaumeGomez:
I'm surprised we missed these cases. Nice catch, thanks!

PR #191, opened 2025-03-28
Fix `get_num_dimensions` error on empty geometries

Oreilles:
`GEOSGeom_getDimensions_r` returns 0 for (Multi)Point, 1 for (Multi)Lines,  2 for (Multi)Polygons,  and -1 for empty GeometryCollection.

Reference:
https://github.com/libgeos/geos/blob/1e4d0b01fe307b347fb629793baf8488146d041b/include/geos/geom/Dimension.h#L27-L47

### DIFF
--- a/src/geometry.rs
+++ b/src/geometry.rs
@@ -1043,7 +1043,7 @@ pub trait Geom: AsRaw<RawType = GEOSGeometry> {
     ///
     /// assert_eq!(geom.get_num_dimensions(), Ok(2));
     /// ```
-    fn get_num_dimensions(&self) -> GResult<usize>;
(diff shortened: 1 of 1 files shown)


src/geometry.rs
```diff
@@ -1043,7 +1043,7 @@ pub trait Geom: AsRaw<RawType = GEOSGeometry> {
     ///
     /// assert_eq!(geom.get_num_dimensions(), Ok(2));
     /// ```
-    fn get_num_dimensions(&self) -> GResult<usize>;
+    fn get_num_dimensions(&self) -> GResult<i32>;
```
GuillaumeGomez, line 1046: If it returns `-1`, it's supposed to be handled by the binding, not by the user.

PR #196, opened 2025-04-02
Add GEOS 3.12 and 3.13 bindings

Oreilles:
Upgrades bundled GEOS version and adds the following functions: relate, relate_pattern, difference_prec, sym_difference_prec, union_prec, unary_union_prec, coverage_union, intersection_prec, clip_by_rect, densify, remove_repeated_points, concave_hull, transform_xy, get_extent, line_merge_directed, create_multicurve, create_rectangle, create_circular_string, create_empty_circular_string, create_empty_compound_curve, create_empty_curve_polygon


GuillaumeGomez:
The CI is unhappy with the changes apparently. :-/

src/enums.rs
```diff
@@ -17,7 +18,8 @@ impl TryFrom<u32> for CoordDimensions {
             1 => Ok(CoordDimensions::OneD),
             2 => Ok(CoordDimensions::TwoD),
             3 => Ok(CoordDimensions::ThreeD),
-            _ => Err("dimensions must be >= 1 and <= 3"),
+            4 => Ok(CoordDimensions::ThreeD),
```
mthh, line 21: Shouldn't this be 
```suggestion
            4 => Ok(CoordDimensions::FourD),
```

Oreilles: Oh, nice catch.

GuillaumeGomez:
Build failure for newer versions.

Oreilles:
The WKTWriter behavior was changed in 3.12 so we'll have to update all the tests output to make that pass.

.github/workflows/CI.yml
```diff
@@ -132,25 +136,30 @@ jobs:
           ccache --max-size=2G --set-config cache_dir=~/.ccache
           ccache --show-stats --set-config cache_dir=~/.ccache
 
+      - name: Install cmake
+        uses: aminya/install-cmake@master
+        with:
+          cmake: 3.31.6
```
jayvdb, line 142: this is unmaintained. See https://github.com/Symbitic/install-cmake/pull/4#issuecomment-2770774769 for recommendation by the author of this action

Oreilles: Thanks for helping with this! I found out that we actually didn't need this after upgrading the GEOS version tags to their latest minor.

src/coord_seq.rs
```diff
@@ -92,9 +92,7 @@ impl CoordSeq {
 
         if size > 0 {
             let dims = data[0].as_ref().len();
-            if let Err(e) = CoordDimensions::try_from(dims as u32) {
-                return Err(Error::GenericError(e.to_owned()));
-            }
+            let _ = CoordDimensions::try_from(dims as u32)?;
```
GuillaumeGomez, line 95: I think:

```suggestion
            CoordDimensions::try_from(dims as u32)?;
```

works fine, no?

src/coord_seq.rs
```diff
@@ -279,7 +277,7 @@ impl CoordSeq {
     /// ```
     /// use geos::{CoordDimensions, CoordSeq};
     ///
-    /// let mut coords = CoordSeq::new(1, CoordDimensions::OneD)
+    /// let mut coords = CoordSeq::new(1, CoordDimensions::TwoD)
```
GuillaumeGomez, line 280: Same question.

src/coord_seq.rs
```diff
@@ -392,7 +390,7 @@ impl CoordSeq {
...
```
GuillaumeGomez, line 393: Same question.

src/coord_seq.rs
```diff
@@ -724,23 +722,23 @@ impl CoordSeq {
     /// ```
     /// use geos::{CoordDimensions, CoordSeq};
     ///
-    /// let coords = CoordSeq::new(2, CoordDimensions::OneD)
+    /// let coords = CoordSeq::new(2, CoordDimensions::ThreeD)
```
GuillaumeGomez, line 725: Isn't it two dimensions?

src/enums.rs
```diff
@@ -68,6 +53,8 @@ impl Into<c_int> for Dimensions {
 pub enum OutputDimension {
     TwoD,
     ThreeD,
+    #[cfg(feature = "v3_12_0")]
+    FourD,
```
GuillaumeGomez, line 57: Was there a `OneD` added for this enum too maybe?

src/coord_seq.rs
```diff
@@ -14,7 +14,7 @@ type AsArrayOutput = (Vec<f64>, Vec<f64>, Option<Vec<f64>>, Option<Vec<f64>>);
 /// ```
 /// use geos::{CoordDimensions, CoordSeq};
 ///
-/// let mut coords = CoordSeq::new(1, CoordDimensions::OneD)
+/// let mut coords = CoordSeq::new(1, CoordDimensions::TwoD)
```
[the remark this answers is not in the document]
Oreilles: One dimensional coordinate sequences are not allowed anymore in 3.12, using `CoordDimensions::TwoD` allows us to have a single test for all versions.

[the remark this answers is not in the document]
GuillaumeGomez: Weird. Well if it works...

src/coord_seq.rs
Oreilles: The 2 on the left is the length of the coordinate sequence, not its dimension. This line just creates a new coordinate sequence with dimension 3, and confirm on the next line that it has been created properly.

src/coord_seq.rs
Oreilles: Not sure clippy would be happy with unused function return values ?

src/enums.rs
Oreilles: This enum is used in the WKT writer, where 1 is not a valid value for the output dimension (only 2, 3, 4 are).

src/coord_seq.rs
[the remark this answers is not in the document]
Oreilles: By the way I'm not even sure why creating a 1-dimensional coordinate sequence is allowed in the first place... It's not like you can create a geometry from it...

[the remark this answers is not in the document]
GuillaumeGomez: No idea. ^^'

PR #198, opened 2025-04-10
Remove most macro rules, so format and lint can run

Oreilles:
This is a refactor of the `Geom` and other `TryFrom` implementation, which remove the need for those hacky macro rules. It provides improved readability and allow formatting and linting to be run within (it wasn't the case before).

In order to achieve this, a new trait `OriginalGeometry` is introduced with a method `original`, which returns `self` for `Geometry` and `self.original` for `ConstGeometry` (maybe naming can be improved on that).

### DIFF
--- a/src/geometry.rs
+++ b/src/geometry.rs
@@ -51,7 +51,29 @@ pub struct ConstGeometry<'a> {
     pub(crate) original: &'a Geometry,
 }
 
-pub trait Geom: AsRaw<RawType = GEOSGeometry> {
+pub trait OriginalGeometry {
+    fn original(&self) -> &Geometry;
+}
+
+impl OriginalGeometry for Geometry {
+    fn original(&self) -> &Geometry {
+        self
+    }
+}
+
+impl OriginalGeometry for ConstGeometry<'_> {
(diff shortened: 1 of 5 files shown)


src/geometry.rs
```diff
@@ -51,7 +51,29 @@ pub struct ConstGeometry<'a> {
...
+    }
+}
+
+impl OriginalGeometry for ConstGeometry<'_> {
+    fn original(&self) -> &Geometry {
+        self.original
+    }
+}
+
+unsafe impl Send for Geometry {}
```
GuillaumeGomez, line 70: Is it safe to implement `Send`  and `Sync` on these two types?

Oreilles: That's not a new introduction, it was already implemented for both 

ttps://github.com/georust/geos/pull/198/files#diff-434e341aaed8371258b2ef24e2c036049c8a56b46d90a2777954ecd6de5afb72L2691-L2692

GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

PR #199, opened 2025-04-24
Add macros to validate every GEOS call (and much more)

Oreilles:
Sorry about the changes salad. Every point can be adressed separately, but splitting all this into different PRs would have been a bit too challenging.  First commit should be the more readable, as the second one involves moves of large chunks of unchanged code.

- Add macros `nullcheck`, `errcheck`, `predicate` to validate every GEOS call. This makes geometry type checking in rust unnecessary, and allows retrieving the actual GEOS error messages for all functions.
- Refactor `Geom` trait to merge declaration and implementation.
- Refactor `ContextHandler`.
- Remove errors `InvalidGeometry`, `GeosFunctionError`, `NoConstructionFromNullPtr`, `VoronoiError`, `NormalizeError`.
- Update `GenericError` and `GeosError` formatting.
- `transform_xy` now return `Result<_, E: From<geos::Error>>`. (This allow using failible functions and have access to the exact error they may throw in the closure)
- `is_valid`, `geometry_type`, `clone` now return `GResult<_>`.
- `ConstGeometry` now holds `PhantomData` instead of reference to original.
- `get_num_dimensions` now return `i32` instead of `usize` (for consistency with `GEOSGeom_getDimensions_r` which can return -1, and uses 0 as the error code).
- `STRtree::iterate` and `STRtree::query` now requires `&mut self` instead of `&self` (for consistency with the C function signature, but is it really necessary or can it stay as it was before ?)

### DIFF
--- a/src/functions.rs
+++ b/src/functions.rs
@@ -1,69 +1,74 @@
-use crate::context_handle::{with_context, PtrWrap};
+use crate::context_handle::with_context;
 use crate::enums::*;
-use crate::error::{Error, GResult, PredicateType};
+use crate::error::{Error, GResult};
 use crate::geometry::Geometry;
 use crate::{AsRawMut, ContextHandle, Geom};
 use geos_sys::*;
 use std::convert::TryFrom;
 use std::ffi::CStr;
 use std::os::raw::c_char;
-use std::str;
+use std::ptr::NonNull;
 
 // We need to cleanup only the char* from geos, the const char* are not to be freed.
 // this has to be checked method by method in geos
 // so we provide 2 method to wrap a char* to a string, one that manage (and thus free) the underlying char*
 // and one that does not free it
-pub(crate) unsafe fn unmanaged_string(raw_ptr: *const c_char, caller: &str) -> GResult<String> {
-    if raw_ptr.is_null() {
-        return Err(Error::NoConstructionFromNullPtr(format!(
-            "{caller}::unmanaged_string",
-        )));
-    }
-    let c_str = CStr::from_ptr(raw_ptr);
-    match str::from_utf8(c_str.to_bytes()) {
-        Ok(s) => Ok(s.to_string()),
-        Err(e) => Err(Error::GenericError(format!(
-            "{caller}::unmanaged_string failed: {e}",
-        ))),
-    }
+pub(crate) unsafe fn unmanaged_string(ptr: NonNull<c_char>) -> GResult<String> {
+    let c_str = CStr::from_ptr(ptr.as_ptr());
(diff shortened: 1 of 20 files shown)


src/functions.rs
```diff
@@ -1,69 +1,79 @@
...
+        })
+    }};
 }
 
-pub(crate) fn check_ret(val: i32, p: PredicateType) -> GResult<()> {
-    match val {
-        1 => Ok(()),
-        _ => Err(Error::GeosFunctionError(p, val)),
-    }
+macro_rules! errcheck {
```
GuillaumeGomez, line 46: Generally speaking, it's better to avoid macros if possible. In this case, you can have two versions of `errcheck`: one where you pass `errval` and the function returned value and the other where you call the first version with `0`.

Then you have one less macro. :)

src/functions.rs
```diff
@@ -44,14 +44,6 @@ macro_rules! nullcheck {
...
 
 macro_rules! errcheck {
-    ($func:ident($ctx:ident.as_raw() $(, $($args:expr),* $(,)?)?)) => {{
-        let result = $func($ctx.as_raw()$(, $($args),*)?);
-        if result == 0 {
-            Err($crate::Error::GeosError((stringify!($func), $ctx.get_last_error())))
-        } else {
-            Ok(result)
-        }
-    }};
```
GuillaumeGomez: Still a macro. :p

Oreilles: Oh, you mean that you'd prefer `nullcheck` to be a function rather than a macro ? I agree that it's preferable to avoid macro wherever possible, the reason I used a macro here is because I can infer the context and the function name directly from the macro arguments.

So it's possible to just write:
```rs
errcheck!(GEOSArea_r(ctx.as_raw(), self.as_raw(), &mut n))?;
```
instead of:
```rs
let ret = GEOSArea_r(ctx.as_raw(), self.as_raw(), &mut n);
errcheck("GEOSArea_r", ctx, 0, ret)?;
```
Which would feel a bit bloaty when used on every geos call.

GuillaumeGomez: Oh I didn't notice that you were stringifying the function name. Then yes, let's keep it as a macro. The simplification is already a nice plus. :+1:

GuillaumeGomez (approved):

GuillaumeGomez:
Letting you merge. :)

Oreilles:
Preferably squashed I assume ?

GuillaumeGomez:
You're the co-maintainer now so it's up to you to decide how you want the commit history to look like. I prefer to have a commit for a specific change so if it doesn't match this and you want to keep following the current scheme, then yes. Otherwise do as you please. ;)

PR #210, opened 2025-11-13
Add GEOSMakeValidParams

joonaspessi:
- [x] I agree to follow the project's [code of conduct](https://github.com/georust/.github/blob/main/CODE_OF_CONDUCT.md).
- [ ] I added an entry to the project's change log file if knowledge of this change could be valuable to users.
  - Usually called `CHANGES.md` or `CHANGELOG.md`
  - Prefix changelog entries for breaking changes with "BREAKING: "
---

Implements parameterized make_valid operation following the same pattern as `BufferParams`. 

The existing make_valid() method uses default parameters with no configuration options. GEOS 3.10+ introduced GEOSMakeValidParams to allow:

  1. Algorithm selection - Structure method often produces better results on strongly invalid geometries
  2. Collapsed geometry handling - Control whether degenerate geometries (e.g., collapsed triangles) are
  preserved


joonaspessi:
CI should pass now

